PR #40, opened 2016-07-04
Use global GLPI_PLUGIN_DOC_DIR to store generated files instead of plugin own dirs

trasher:
Also fix a minor bug on class existence check.


orthagh:
It seems great.
Tests & merge incoming (surely friday)

orthagh:
After review, i think we need a part in update script to remove old files.

trasher:
You're right, I did not take care of the update part... We need to re-generate them because tamplates has changed, and remove old ones.

I'll do this change asap.

PR #55, opened 2016-09-20
Fixed #54

### LINKED ISSUE #54: Entity restriction issue
Hello
Let's considere this entity schema 👍 
Root
+---Entity1
+---Entity2

Let's define a block visible in Entity1 (recursively, but doesn't matters) with some field values, visible for a certain type of object : ex Computer
There is an issue with the entity of a block 
The Block tab should only be visible in the Entity where it is defined (in this sample Entity1): computers in entity 1 should have the tab defined previously and not in Entity2

Hope I'm clear enough ;)

tomolimo:
Added a test to restrict entity, to be sure that object entity permits the current container(s)


tomolimo:
At the end it is ok,
Could you push this commit?
Thank you,
Regards,
Tomolimo

PR #92, opened 2016-11-29
Remove table when class does not exists (should not happen... but...)

### LINKED ISSUE #90: Improve uninstall
When trying to uninstall the plugin from a bogus state, it fails if some classes are missing; it should not be a blocker on uninstalling.

trasher:
fixes #90


trasher:
Thanks for the review :)

PR #99, opened 2016-12-12
Add labels translations; closes #88

### LINKED ISSUE #88: Translate fields

trasher:


front/labeltranslation.form.php
```diff
@@ -0,0 +1,50 @@
...
+
+ You should have received a copy of the GNU General Public License
+ along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ --------------------------------------------------------------------------
+ */
+
+
+/** @file
+ * @brief
+ * @since version 0.85
```
orthagh: This header in wrong (part of glpi, since 0.85).
Copy/paste i think, the plugin may need a clean in its headers

inc/field.class.php
```diff
@@ -121,16 +121,11 @@ function prepareInputForAdd($input) {
 
       return $input;
    }
-   /*function prepareInputForUpdate($input) {
```
orthagh: part removed  ?

inc/field.class.php
```diff
@@ -121,16 +121,11 @@ function prepareInputForAdd($input) {
...
-      //parse name
-      //$input['name'] = $this->prepareName($input);
-
-      return $input;
-   }*/
 
    function pre_deleteItem() {
       global $DB;
 
+      //TODO: remove labels translations
```
orthagh: TODO not filled

inc/field.class.php
trasher: From a while apparently: e6b0b38b3d3d01d811b073eaf72bb06a23b6218f

inc/field.class.php
trasher: Good catch, I've finally missed this one :)

front/labeltranslation.form.php
trasher: Done for this file. Other headers are out of the scope of this PR; see #100

inc/field.class.php
trasher: Done. Also for containers.

PR #102, opened 2016-12-13
fix add new field

orthagh:
with the merge of #99, the form for field addition fails


orthagh:
Still some issues in field addition, currently investigating

orthagh:
ok fix done

PR #108, opened 2016-12-14
Add user preferences support; fixes #5

### LINKED ISSUE #5: Allow tab on user preferences
It would be great to allow adding a tab in user's preferences

trasher:
This one needs a few hacks, since the Preference object does not inherit CommonDBTM


trasher:
Will have to be reworked when hook will be used; this would change a lot :)

PR #118, opened 2016-12-20
Hardcode domtab possible values

trasher:
Limit proposed values in domtab feature to usable ones only


trasher:
Thanks for the review :)

PR #135, opened 2017-01-06
Use global GLPI standards (version 0.5)

trasher:


trasher:
Thanks for the review. I do not understand what is "disturbing"

orthagh:
For a coding standard pr & review, you have a method parameters order switch

trasher:
Yes, this is because of standards: parameters with default values must be at the end ;)

PR #139, opened 2017-01-11
Feature/atoum

trasher:


trasher:
Because 2.8 is not PHP 7.1 compliant. The applied patch has been retrieved from upstream 3.0 branch.

orthagh:
ok for me so :)

trasher:
Thanks!

PR #140, opened 2017-01-11
Build against various GLPI versions

trasher:
Also, do not retrieve full history while cloning


trasher:
Réintegrated and modified in #119

PR #155, opened 2017-02-06
Add robo CS task; fix cs on JS file, use task for travis

trasher:


trasher:
Thansk for the review; but this PR is outdated since project now use external tools

PR #161, opened 2017-02-20
hotfix/1.5.1

orthagh:
should fix #112

PR #162, opened 2017-02-27
Tomolimo patch to normalize backslashes

tomolimo:
In the realpath( ) test


orthagh:
Also ok for me.
Will merge into #161

PR #164, opened 2017-02-28
Drop Zend\Loader usage; use Fedora\Autoloader instead

### LINKED ISSUE #163: rework autoloader
see https://github.com/glpi-project/glpi/pull/1855

we need to include (or drop) Zend\Loader\SplAutoloader

trasher:
Use namespaces for generated classes so we can use PSR4

Fixes #163


orthagh:
will merge into #161

PR #174, opened 2017-03-09
Hotfix/1.6.1

trasher:


trasher:
I do as well :'(

I had solutions for parts of the problem (overriding some core ajax calls); but overriding is not always possible (because of the way requires/includes are done).

We would globally have to make changes on the way the `itemtypes` are retrieved; because of the automatic sanitize of all _POST and _GET :-/

PR #195, opened 2017-07-11
Update to latest atoum

trasher:
Use core's DbTestCase

### DIFF
--- a/tests/FieldsDbTestCase.php
+++ b/tests/FieldsDbTestCase.php
@@ -32,7 +32,7 @@
 
 // Generic test classe, to be extended for CommonDBTM Object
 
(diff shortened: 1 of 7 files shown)


tests/FieldsDbTestCase.php
```diff
@@ -32,7 +32,7 @@
 
 // Generic test classe, to be extended for CommonDBTM Object
 
-class DbTestCase extends atoum {
+class FieldsDbTestCase extends \DbTestCase {
```
orthagh: great idea

trasher: :) Thanks

PR #199, opened 2017-08-14
correction default value for type yesno

lermit4:
because the value of no is 0


orthagh:
Hello @lermit4 .

In pr #203, i pushed a better fix for this issue:
e157078

PR #203, opened 2017-08-31
1.7.0

orthagh:


inc/container.class.php
```diff
@@ -1408,4 +1451,12 @@ private static function getSubtypes($item) {
 
       return $tabs;
    }
+
+   static function getClassname($itemtype = "", $raw_name = "") {
```
trasher: Missign phpdoc

inc/container.class.php
```diff
@@ -1408,4 +1451,12 @@ private static function getSubtypes($item) {
...
+      return "PluginFields".ucfirst(self::getSystemName($itemtype, $raw_name));
+   }
+
+   static function getSystemName($itemtype = "", $raw_name = "") {
```
trasher: Missing phpdoc

inc/field.class.php
```diff
@@ -348,10 +341,10 @@ function showSummary($container) {
 
                echo '<td class="rowhandler control center">';
                echo '<div class="drag row" style="cursor:move;border:none !important;">';
-               echo '<img src="../pics/drag.png" alt="#" title="' . __('Move') .'" width="16" height="16" />';
+               echo '<img src="../pics/drag.png" alt="#" title="'.__('Move').'" width="16" height="16">';
```
trasher: Standalone markup is generally closed ;)

orthagh: in html5, it's optional

trasher: Yes it is... But it is also considered as a good practice to close all tags. Well this is not important.

trasher:
Sounds OK, I've to test the numerous possibilities...

I'm not sure removing the OS stuff is the correct way to go; even if this is a quite huge change :/

PR #278, opened 2018-08-30
Develop

KevinStephane:
Hello,
Is it possible to add a counter in the tab when these values ​​are filled in a field.
Thank you


KevinStephane:
Bonjour, 
Pourquoi fermer ma demande ? :-)
Merci

PR #364, opened 2020-02-19
add operating system dropdown

nenes82:
Add operating system  dropdown

bugfixes: When all tabs are displayed, the fields added to the dom are not displayed


tomolimo (approved):
It seems ok for me.

PR #425, opened 2021-01-13
Fix extra fields display on creation form

AdrienClairembault:
Internal ref : !21319.

Extra fields are missing on creation forms if the user's current tab for this itemtype (`$_SESSION['glpi_tabs']`) is not the main one ('$main').

Fix seems to be to add a special case to cover this case like it was done before for the 'showPrimaryForm' and 'showFormHelpdesk' functions.

### DIFF
--- a/inc/field.class.php
+++ b/inc/field.class.php
@@ -544,6 +544,7 @@ class PluginFieldsField extends CommonDBTM {
 
       $subtype = isset($_SESSION['glpi_tabs'][strtolower($item::getType())]) ? $_SESSION['glpi_tabs'][strtolower($item::getType())] : "";
       $type = substr($subtype, -strlen('$main')) === '$main'
+              || in_array('showForm', $functions)
               || in_array('showPrimaryForm', $functions)
               || in_array('showFormHelpdesk', $functions)
                ? 'dom'


cedric-anne (approved):

orthagh (approved):

AdrienClairembault:
Can be merged & released, both internal ref confirmed that the patch fixed the issue on their sides too.

tomolimo:
Hello @cedric-anne 
This fix is not only fixing for creation, but also for edition.
I've backported this fix on our current version which is running on a 9.5.3 GLPI release, and it fixes an issue we had.

To reproduce:

1. login into GLPI
2. go to ticket list
3. Open in a new browser page a ticket, and go to this page: "processing ticket" tab open by default, click on 'Ticket' tab, click on 'processing ticket' tab and Reload the page (F5).
4. keep this page open
5. go back to ticket list, open in a another browser page another ticket (from the ticket list), and go to this newly open page: "processing ticket" tab will open by default (current tab in session), 
6. click on main tab (='Ticket' tab): --> the custom fields don't appear...

Thank you very much for this fix

Regards,
Tomolimo

PR #461, opened 2021-11-04
Fix path check

AdrienClairembault:
If you have a mounted disk, the path check will never succeed because the variables will look like this:
```php
$realpath = /mnt/diskhome/home/...
$frontpath = /home/...
```
It seems like `realpath` should be applied to both path we are trying to compare.


cedric-anne (approved):

AdrienClairembault:
Found another front file that needed the same change.

cedric-anne (approved):

PR #466, opened 2022-01-14
fix(core): better check to include file from path

stonebuzz:
From CLOUD context ```RuntimeException``` is throw when we try to go to dropdowns created from fields plugin 

```realpath``` =
 ```/AAA/BBB/CCC/DDD/EEE/glpi_files/_plugins/fields/front/secteurdaffectationfielddropdown.php```

```frontpath``` = 
```/XXX/YYY/CCC/DDD/EEE/glpi_files/_plugins/fields/front```

To prevent load missing or unreadable file just check if it readable or if exist

Internal 23215

### DIFF
--- a/front/commondropdown.form.php
+++ b/front/commondropdown.form.php
@@ -1,11 +1,8 @@
 <?php
 
 include "../../../inc/includes.php";
-$path = PLUGINFIELDS_FRONT_PATH . '/' . $_REQUEST['ddtype'] . '.form.php';
-$realpath = str_replace( "\\", "/", realpath($path));
-$frontpath = str_replace("\\", "/", realpath(PLUGINFIELDS_FRONT_PATH));
-if (strpos($realpath, $frontpath) === 0) {
-    include_once $path;
-} else {
-    throw new \RuntimeException('Attempt to load unsecure or missing ' . $path .'!');
+if (preg_match('/[a-z]/i', $_REQUEST['ddtype']) !== 1) {
+   throw new \RuntimeException(sprintf('Invalid itemtype "%"', $_REQUEST['ddtype']));
 }
+$path = PLUGINFIELDS_FRONT_PATH . '/' . $_REQUEST['ddtype'] . '.form.php';
+require_once $path;
--- a/front/commondropdown.php
+++ b/front/commondropdown.php
@@ -1,11 +1,8 @@
 <?php
 
 include "../../../inc/includes.php";
-$path = PLUGINFIELDS_FRONT_PATH . '/' . $_REQUEST['ddtype'] . '.php';
-$realpath = str_replace( "\\", "/", realpath($path));
-$frontpath = str_replace("\\", "/", realpath(PLUGINFIELDS_FRONT_PATH));
-if (strpos($realpath, $frontpath) === 0) {
-    include_once $path;
-} else {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: front/commondropdown.php [hunk 1]
```diff
@@ -2,10 +2,9 @@
 
 include "../../../inc/includes.php";
 $path = PLUGINFIELDS_FRONT_PATH . '/' . $_REQUEST['ddtype'] . '.php';
```
cedric-anne, line 4: This test was made tio ensure that this script is not used to execute code that is located outside the plugin directory.

```suggestion
if (preg_match('/[a-z]/i', $_REQUEST['ddtype']) !== 1) {
   throw new \RuntimeException(sprintf('Invalid itemtype "%"', $_REQUEST['ddtype']));
}
$path = PLUGINFIELDS_FRONT_PATH . '/' . $_REQUEST['ddtype'] . '.php';
```

cedric-anne:
`commondropdown.form.php` should be updated too.

front/commondropdown.php
AdrienClairembault: Seems like a good way to do it :+1:
If understand correctly you check that the file name contains only text so that you can't go to another folder with `../../..`.

cedric-anne (approved):

AdrienClairembault (approved):

front/commondropdown.form.php
```diff
@@ -1,11 +1,8 @@
...
-$frontpath = str_replace("\\", "/", realpath(PLUGINFIELDS_FRONT_PATH));
-if (strpos($realpath, $frontpath) === 0) {
-    include_once $path;
-} else {
-    throw new \RuntimeException('Attempt to load unsecure or missing ' . $path .'!');
+if (preg_match('/[a-z]/i', $_REQUEST['ddtype']) !== 1) {
+   throw new \RuntimeException(sprintf('Invalid itemtype "%"', $_REQUEST['ddtype']));
 }
+$path = PLUGINFIELDS_FRONT_PATH . '/' . $_REQUEST['ddtype'] . '.form.php';
+require_once $path;
```
orthagh, line 8:
```suggestion
require_once $path;

```

orthagh (changes requested):

front/commondropdown.php
```diff
@@ -1,11 +1,9 @@
...
-if (strpos($realpath, $frontpath) === 0) {
-    include_once $path;
-} else {
-    throw new \RuntimeException('Attempt to load unsecure or missing ' . $path .'!');
+if (preg_match('/[a-z]/i', $_REQUEST['ddtype']) !== 1) {
+   throw new \RuntimeException(sprintf('Invalid itemtype "%"', $_REQUEST['ddtype']));
 }
+$path = PLUGINFIELDS_FRONT_PATH . '/' . $_REQUEST['ddtype'] . '.php';
+require_once $path;
+
```
cedric-anne, line 9:
```suggestion
```

front/commondropdown.form.php
AdrienClairembault: @stonebuzz There is an option in vscode to do it automatically btw:
image

orthagh (approved):

front/commondropdown.form.php
stonebuzz: :+1:

PR #525, opened 2022-06-01
Add command to fix incorrectly deleted fields from containers tables

AdrienClairembault:
Internal ref: !24075.

EDITED after #526 (see comments)

An issue affected field removal in 1.15.0, 1.15.1 and 1.15.2.
Using these versions, removing a field from a container would drop the field from glpi_plugin_fields_fields but not from the custom container table.

This command remove these incorrectly deleted fields.

Command examples:
image

Unrelated changes kept from the previous iteration:
* Add `PluginFieldsContainerTemplate` base class for custom containers classes in order to reduce the generated php classes for each containers.

The main advantage is that you can edit the generic code directly from `PluginFieldsContainerTemplate` and it will impact all custom container classes without having to "regenerate" the PHP files of these custom classes.

### DIFF
--- a/inc/migration.class.php
+++ b/inc/migration.class.php
@@ -28,96 +28,227 @@
  * -------------------------------------------------------------------------
  */
 
-class PluginFieldsMigration extends Migration {
+class PluginFieldsMigration extends Migration
+{
+    public function __construct($ver = "")
+    {
+        parent::__construct($ver);
+    }
 
-   function __construct($ver = "") {
-      parent::__construct($ver);
-   }
+    public static function install(Migration $migration, $version)
+    {
+        global $DB;
 
-   static function install(Migration $migration, $version) {
-      global $DB;
+        $fields_migration = new self();
 
-      $fields_migration = new self;
+        if ($DB->tableExists("glpi_plugin_customfields_fields")) {
+            if (!$fields_migration->updateFromCustomfields()) {
+                return false;
+            }
+        }
 
-      if ($DB->tableExists("glpi_plugin_customfields_fields")) {
-         if (!$fields_migration->updateFromCustomfields()) {
-            return false;
-         }
-      }
+        return true;
+    }
 
-      return true;
-   }
+    public static function uninstall()
+    {
+        return true;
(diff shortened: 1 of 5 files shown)


cedric-anne:
Even if such a solution would work, I think it would be better to delete data from DB when field/dropdown is deleted, for instance in `PluginFieldsField::pre_deleteItem()` and in `PluginFieldsDropdown::destroy()`.
For existing data, cleaning could be handled in migration.

AdrienClairembault:
Actually, after re-reading the code I've found that deletion was just broken since 1.15.0.

`removeField` was called with the wrong parameters since https://github.com/pluginsGLPI/fields/commit/954864bbd0b8817f9a7815f6447c49fa9d6cbd95#diff-8fd65f1eb4cf4f2c2b956a535e6de83c36b63bfabfd1a3c9131b99a39e47b9a7L219

inc/migration.class.php
```diff
@@ -28,96 +28,227 @@
...
-
-      $default_key_sign = DBConnection::getDefaultPrimaryKeySignOption();
-
-      $fields = [];
-      switch (true) {
-         case $field_type === 'header':
-            // header type is for read-only display purpose only and has no SQL field
-            break;
-         case $field_type === 'dropdown':
-         case preg_match('/^dropdown-.+/i', $field_type):
```
cedric-anne, line 98: `case preg_match('/^dropdown-.+/i', $field_type):` was a legitimate case and should be preserved.

Diff is huge here, and it is hard to validate the code. Unless I am wrong, if #528 is merged, only the new command and corresponding `PluginFieldsMigration` methods are required.

I would prefer serarate PR for PSR-12 migration and for migration of PHP template files to abstract classes (should be done globally). We could work on these later.

AdrienClairembault: Ok, let's merge #528 and then I'll open separate PR for each topics.

PS: for the huge diff, it's mostly whitespaces and you can ignore them:
image

AdrienClairembault:
Replaced by #532.

PR #532, opened 2022-06-03
Command: fix dropped fields

AdrienClairembault:
Internal ref: !24075.

An issue affected field removal in 1.15.0, 1.15.1 and 1.15.2.
Using these versions, removing a field from a container would drop the field from glpi_plugin_fields_fields but not from the custom container table.

This command remove these incorrectly deleted fields.

Examples:
image

### DIFF
--- a/inc/checkdatabasecommand.class.php
+++ b/inc/checkdatabasecommand.class.php
@@ -34,65 +34,72 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
-class PluginFieldsFixDroppedFieldsCommand extends AbstractCommand
+class PluginFieldsCheckDatabaseCommand extends AbstractCommand
 {
     protected function configure()
     {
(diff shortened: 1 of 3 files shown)


AdrienClairembault:
Up @cedric-anne

cedric-anne (approved):

cedric-anne:
I adapted command to make it more generic:
 - renamed to `check_database`
 - change option `delete` to `fix`
 - move bug details in help section

I added some colors to command ouput
image

AdrienClairembault:
All good for me, great work 👍 

I'm just not sure about "Database contains 4 gone fields" it sound of bit weird maybe ?
I think any of the following options would be a better translation :
- "The database contains 4 incorrectly deleted fields"
- "The database contains 4 deleted fields"
- "The database contains 4 missing fields"

Maybe @cconard96 can help us here ;)

cconard96:
All of those options make it sound like something was deleted that shouldn't have been when instead, there is orphaned database fields.
It may be better to say "The database contains orphaned data from 4 previously deleted fields" or change "previously" to "improperly".

PR #538, opened 2022-06-13
fix(searchoption): search number as text

### LINKED ISSUE #536: Searches doesn't work on fields created as number on 1.15.3
**Describe the bug**
After migrate from version 1.12.10 to 1.15.3.
When you try to search (ex. Software items), filter by a field created as number, cannot find items.

**To Reproduce**

Go to 'Setup' / 'Additional Fields'
Click on 'Add'
  (create a container for any item type)
Add new Field as Number
Complete on an item this field with any number.
The make a search using filter this field and using "contain" write that number

**Expected behavior**
I expect to view the item found in this search.

Screenshots
image

image

But the item exits.

image

Additional Comments
Using Debug mode, I can saw that the SQL Sentence is appended a "." after expression to search:

If I run that sentence on SQL query manager, also have any results. But If I remove the "." can obtain the item that I looking for.

`glpi_plugin_fields_softwaresoftwares_otroreferentefield`.`itemtype` = 'Software' )
WHERE `glpi_softwares`.`is_deleted` = 0 AND `glpi_softwares`.`is_template` = 0 AND ( (`glpi_plugin_fields_softwaresoftwares_idaplicacinfield`.`idaplicacinfield` LIKE '%560.%' ) ) GROUP BY `glpi_softwares`.`id`
ORDER BY `id`

GLPI / Plugins (please complete the following information):

GLPI version 10
list of all plugins with their version (available in GLPI > Setup > General > tab System > bottom of the page)
fields Name: Additional Fields Version: 1.15.3 State: Enabled

thanks in advance!

OS: Ubuntu 20.04 LTS
Browser chrome and safari have the same issues
Version Latest for both.

stonebuzz:
Fix #536 , internal ref !24208

@cconard96 what do you think about this fix?

Best regards

### DIFF
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -1631,10 +1631,8 @@ HTML;
                     $opt[$i]['datatype'] = "bool";
                     break;
                 case 'textarea':
-                    $opt[$i]['datatype'] = "text";
-                    break;
                 case 'number':
-                    $opt[$i]['datatype'] = "decimal";
+                    $opt[$i]['datatype'] = "text";
                     break;
                 case 'date':
                 case 'datetime':


cconard96:
As a short term fix, this seems OK. In the long term, using the correct column type would be ideal.

cconard96 (approved):

PR #616, opened 2022-12-26
Do not use REPLACE INTO

cconard96:
Remove use of non-standard SQL `REPLACE INTO` raw query and replace with `DBmysql::updateOrInsert`.


cedric-anne (commented):
Maybe `Migration::updateDisplayPrefs()` could be used here.

PR #643, opened 2023-03-20
fix(core): escape data during field update

stonebuzz:
Fix update process.
escape data during field update

### DIFF
--- a/inc/toolbox.class.php
+++ b/inc/toolbox.class.php
@@ -28,6 +28,8 @@
  * -------------------------------------------------------------------------
  */
 
+use Glpi\Toolbox\Sanitizer;
+
 class PluginFieldsToolbox
 {
    /**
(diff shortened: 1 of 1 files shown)


DooBio:
Sorry I made the change but no change.

I still emptied the cache and deactivated / activated the plugin

Sorry

` */
 
use Glpi\Toolbox\Sanitizer;

class PluginFieldsToolbox
{
   /**
    * Get a clean system name from a label.
    *
    * @param string $label
    *
    * @return string
    */
--------
        foreach ($fields as $field) {
            $field_copy = $field;
            unset($field_copy['name']);
            if ($field['name'] !== (new PluginFieldsField())->prepareName($field_copy, false)) {
                $bad_named_fields[] = Sanitizer::dbEscapeRecursive($field);
            }
        }
`

cedric-anne (approved):

PR #670, opened 2023-07-05
HL API integration

cconard96:


SkiAn2007:
Would be nice to have ability to get\update custom fields via new API

PR #676, opened 2023-07-20
Fix yes/no default value

Aeto-J:
For yesno field default value must be 0

If default value is not set and readonly is active that create an sql error :

Error: Incorrect integer value: '' for column 'fieldname' at row 1


tomolimo:
Hello,
For me it's OK to have this default value added or yesno fields
Thank you
Regards,
Tomolimo

tomolimo (approved):

cedric-anne (approved):

PR #704, opened 2023-09-07
Fix issue with wrong default value for container

tomolimo:
Default value for $id must be false, otherwise the searches for container in preItem() will failed.

### DIFF
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -1532,7 +1532,7 @@ HTML;
 
         $container = new PluginFieldsContainer();
         $itemtypes = $container->find($condition);
-        $id = 0;
+        $id = false; // default value when no container matches criterion
         if (count($itemtypes) < 1) {
             return false;
         }


stonebuzz (approved):

stonebuzz:
Hello @tomolimo , could you nevertheless explain the problem you encountered.

I imagine a type/value comparison problem (===)

https://github.com/pluginsGLPI/fields/blob/264ea43575da66a86fd853e8a260b32a31c4c646/inc/container.class.php#L1628-L1633

and here

https://github.com/pluginsGLPI/fields/blob/264ea43575da66a86fd853e8a260b32a31c4c646/inc/field.class.php#L872-L876

tomolimo:
Hello @stonebuzz,

Yes, you are right: the root cause of this issue is here:

image

If ‘dom’ test returns 0 when there is no ‘dom’ container, then it doesn’t search for ‘tab’ containers.

Thank you,
Regards,
Tomolimo

PR #741, opened 2023-12-15
Fix(Container): deactivate domtab that no longer works and stop offer…

stonebuzz:
Since GLPI 10

```Ticket``` ```Solution``` tab no longer exist
```Problem``` ```Solution``` tab no longer exist
```Change``` ```Analysis``` / ```Solution``` / ```Plans``` no longer exist

So, deactivate related ```container``` (with type ```Insertion in the form of a specific tab (before save button)```) and remove the ability to create containers on these tabs (tabs on ```Entity``` are maintained as they are still functional))

**Release note should mention this :**

Adding a container (with type ```Insertion in the form of a specific tab (before save button)```) to the following tabs is no longer possible (as it no longer works since version 10)
- Change -> analysis
- Change -> plans
- Change -> solution
- Ticket -> solution
- Problem -> solution

All related container have been deactivated

For the ```Solution```, it is now possible to add a block on ITILSolution, only for the "Insertion in the form (before save button)" type.

**IMPORTANT**
no migration proposed from the old type to the new one, 
as it hasn't worked for a while, I don't suppose anyone has added a block 
and required lot of work

 
Fix : 
https://github.com/pluginsGLPI/fields/issues/737
https://github.com/pluginsGLPI/fields/issues/407 
https://github.com/pluginsGLPI/fields/issues/599 
https://github.com/pluginsGLPI/fields/issues/456

### DIFF
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -898,11 +898,17 @@ HTML;
     public static function showFormItemtype($params = [])
     {
         $is_domtab = isset($params['type']) && $params['type'] == 'domtab';
+        $values = self::getItemtypes($is_domtab);
+
+        //remove ITISolution from values if type is tab
+        if (!isset($params['type']) || (isset($params['type']) && $params['type'] == 'tab')) {
+            unset($values[__('Assistance')]['ITILSolution']);
+        }
 
         $rand = $params['rand'];
         Dropdown::showFromArray(
             "itemtypes",
-            self::getItemtypes($is_domtab),
+            $values,
             [
                 'rand'                => $rand,
                 'multiple'            => !$is_domtab,
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -323,8 +323,9 @@ class PluginFieldsContainer extends CommonDBTM
             }
         }
 
-        //Ticket Solution tab no longer exist move to ITILSolution dom
-        //Problem Solution tab no longer exist move to ITILSolution dom
+
+        //Ticket Solution tab no longer exist exist disable it
+        //Problem Solution tab no longer exist exist disable it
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Apashh** [comment]
Hello,

If i understand, now its not possible to add a field on solution ? Like this ?
image

A solution have been created : https://github.com/pluginsGLPI/fields/pull/612
But not work

Can you explain your branch and your code ? What is the solution to use a field before send solution ?

Thanks !

**Comment by stonebuzz** [comment]
> For the `Solution`, it is now possible to add a block on ITILSolution, only for the "Insertion in the form (before save button)" type.


For the Solution, it is now possible to add a block on ITILSolution, only for the "Insertion in the form (before save button)" type.

**Comment by Apashh** [comment]
> > For the `Solution`, it is now possible to add a block on ITILSolution, only for the "Insertion in the form (before save button)" type.
> 
> 
> 
> 
> 
> For the Solution, it is now possible to add a block on ITILSolution, only for the "Insertion in the form (before save button)" type.
> 
> 
> 
> 

Okay its not a good solution, it will be better to add field directely on solution reply.. So shame...

File: inc/container.class.php [hunk 1]
```diff
@@ -323,6 +323,27 @@ public static function installUserData(Migration $migration, $version)
             }
         }
 
+
+        //Ticket Solution tab no longer exist exist disable it
+        //Problem Solution tab no longer exist exist disable it
```
Rom1-B, line 328:
```suggestion
        //Ticket Solution tab no longer exist disable it
        //Problem Solution tab no longer exist disable it
```

Rom1-B (changes requested):

Rom1-B (approved):

AdrienClairembault (approved):

PR #798, opened 2024-06-19
fix: Prevent error when classname is null

Rom1-B:
Prevent this error message:

```
glpiphplog.CRITICAL:   *** Uncaught Exception TypeError: get_parent_class(): Argument #1 ($object_or_class) must be an object or a valid class name, null given in /var/www/glpi/files/_plugins/fields/inc/computerchampsadditionnelordinateurinjection.class.php at line 15
  Backtrace :
  ...ampsadditionnelordinateurinjection.class.php:15 get_parent_class()
  ...e/datainjection/inc/injectiontype.class.php:147 PluginFieldsComputerchampsadditionnelordinateurInjection::getTable()
  ...etplace/datainjection/inc/mapping.class.php:145 PluginDatainjectionInjectionType::dropdownLinkedTypes()
  marketplace/datainjection/inc/model.class.php:843  PluginDatainjectionMapping::showFormMappings()
  src/CommonGLPI.php:694                             PluginDatainjectionModel::displayTabContentForItem()
  ajax/common.tabs.php:120                           CommonGLPI::displayStandardTab()
  public/index.php:82                                require()

```

### DIFF
--- a/templates/injection.class.tpl
+++ b/templates/injection.class.tpl
@@ -2,77 +2,86 @@
 
 class %%CLASSNAME%%Injection extends %%CLASSNAME%% implements PluginDatainjectionInjectionInterface
 {
-   static $rightname = 'plugin_datainjection_model';
+    static $rightname = 'plugin_datainjection_model';
 
-   /**
-    * Return the table used to stor this object
-    *
-    * @see CommonDBTM::getTable()
-    *
-    * @return string
-   **/
-   static function getTable($classname = null) {
-      return getTableForItemType(get_parent_class($classname));
-   }
+    /**
+     * Return the table used to stor this object
+     *
+     * @see CommonDBTM::getTable()
+     *
+     * @return string
+    **/
+    static function getTable($classname = null) {
+        if ($classname === null) {
+            return '';
+        }
+        $parent_class = get_parent_class($classname);
 
-   static function getTypeName($nb = 0) {
-      return %%ITEMTYPE%%::getTypeName() . " - %%CONTAINER_NAME%%";
-   }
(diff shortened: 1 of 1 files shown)


trasher:
```
        if ($classname === null) {
            return '';
        }
        $parent_class = get_parent_class($classname);

        if ($parent_class === false) {
            return '';
        }
```

could probably be rewritten to:
```
        if (get_parent_class($classname ?? '') === false) {
            return '';
        }
```

AdrienClairembault:
> I do not know well the plugin; but is really a call with `null` as classname legit? I mean, is the case an empty string is returned from that method handled?

I feel the same way, returning an empty string is dangerous here because some other method using this returned value might assume this is a valid table and try to run some operation on it (thus triggering an error elsewhere).

If the core `getTableForItemType` method had support for invalid parameter by returning something like `null`, `''`, `false` or an exception then we could have copied it here.
But since it does not I think the correct fix is to make sure `getTableForItemType` is not called with an invalid parameter.

AdrienClairembault (changes requested):
See comment.

AdrienClairembault (approved):
Nvm, it looks good to me because we are redefining `getTable` here (I read too fast and thought we were redefining `getTableForItemType`).

The core `getTable` from `CommonDBTM` does output `''` is the table can't be found, so this behavior is good to me.

trasher:
> The core `getTable` from `CommonDBTM` does output `''` is the table can't be found, so this behavior is good to me.

Good catch, OK for me then.

trasher (approved):

PR #800, opened 2024-06-25
1.21.10

Rom1-B:


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.21.10] - 2024-06-11
+
+### Fixed
+
+- Prevent error (with Datainjection plugin) when classname is null (#798)
```
AdrienClairembault, line 12: Seems a bit technical, in my opinion if you have more context it would be cleared if it could be explained from a user perspective ("Prevent error while doing XXX action") instead of a technical explanation.

Besides that OK for merge.

AdrienClairembault (approved):

PR #803, opened 2024-07-01
(fix) Update item date_mod

Lainow:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | !33527

After modifying additional fields, the modification date is not updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [unrelease] -
+
+### Fixed
+
+- Update item date_mod after updating additional fields
+
 ## [1.21.10] - 2024-06-11
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-- Update item date_mod after updating additional fields
+- Update main item ```date_mod``` after updating additional fields
 
 ## [1.21.10] - 2024-06-11
 
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -1658,13 +1658,6 @@ HTML;
         return true;
     }
 
-    public static function postItemUpdate(CommonDBTM $item)
-    {
-        echo 'test';
-
-        return true;
-    }
-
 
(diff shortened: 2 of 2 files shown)


stonebuzz (changes requested):

inc/container.class.php
```diff
@@ -1649,6 +1649,7 @@ public static function preItemUpdate(CommonDBTM $item)
                 count($data) == 0
                 || $container->updateFieldsValues($data, $item->getType(), isset($_REQUEST['massiveaction']))
             ) {
+                $item->input['date_mod'] = date("Y-m-d H:i:s");
```
stonebuzz, line 1652:
```suggestion
                $item->input['date_mod'] = $_SESSION["glpi_currenttime"];
```

stonebuzz (changes requested):

AdrienClairembault (commented):
This is weird because GLPI core is already supposed to do it:

image

stonebuzz:
@Lainow can you complete the CHANGELOG ?

stonebuzz:
> This is weird because GLPI core is already supposed to do it:
> 
> image

only for objects which have a 'date_mod' field and which are affected by the update.

In this case, it is the data from the fields plugin that is updated and not the linked GLPI object (so ```$this->updates``` contains nothing for the GLPI object).

AdrienClairembault:
> > This is weird because GLPI core is already supposed to do it:
> > image
> 
> only for objects which have a `date_mod` field and which are affected by the update.
> 
> In this case, it is the data from the fields plugin that is updated and not the linked GLPI object (so `$this->updates` contains nothing for the GLPI object).

I see, thanks for this details.
I've actually had the same issue on forms when updating a linked question and used the same solution so ok for me 👍.

AdrienClairembault (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [unrelease] -
+
+### Fixed
+
+- Update item date_mod after updating additional fields
```
stonebuzz, line 12:
```suggestion
- Update main item ```date_mod``` after updating additional fields
```

stonebuzz (changes requested):

stonebuzz (approved):

PR #807, opened 2024-07-03
Fix(Injection): fix datainjection error with fields

stonebuzz:
revert https://github.com/pluginsGLPI/fields/pull/798

Just ```getTableForItemType``` with ```get_parent_class()``` (without pass ```$classname``` as parameter because is always ```NULL```)

fix : !33438 and !33460 and !33597


stonebuzz:
waiting for customer feedback

AdrienClairembault (approved):

PR #812, opened 2024-07-15
Fix mandatory dropdown fields

RomainLvr:
Related ticket : 33594

Plugins : **Generic Objects** X **Fields**

For a generic `Car` object, for example, if you add multiple mandatory fields pointing to GLPI objects (Dropdowns Groups, Users, etc.), object creation fails with an error indicating that mandatory fields are empty.

image
image

What's more, once this problem had been resolved, it was possible to enter empty values in mandatory fields, because in some cases, if the field in question was not detected in the code (in the case of an empty value) the value was retrieved from the database to check that a mandatory field was indeed non-empty. This meant that an empty value could only be entered in a mandatory field once (which is already too often).

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-- Fix generic object bloc fields mandatory
+- Fix handling of empty mandatory fields in generic objects.
 
 ## [1.21.11] - 2024-07-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [unrelease] -
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


stonebuzz:
the customer seems satisfied
I'm waiting for feedback on further tests of the plugin's overall performance.

(this part of the code is sensitive in my opinion)

Rom1-B (approved):

stonebuzz (approved):

stonebuzz:
@RomainLvr 
can you adapt CHANGELOG.md ?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [unrelease] -
 
+### Fixed
+
+- Fix generic object bloc fields mandatory
```
Rom1-B, line 12:
```suggestion
- Fix handling of empty mandatory fields in generic objects.
```

PR #819, opened 2024-09-04
Fix(Core): fix bulk update for dropdown used in more than one container

stonebuzz:
In fields, a dropdown field can be reused in several containers

However, when several containers use this type of field, it is no longer possible to modify it using mass actions.

The format of the query before the change resulted in an `IN` clause which returned both containers (KO), now we have a `LIKE` clause (which returns a single container (OK)).

### DIFF
--- a/inc/field.class.php
+++ b/inc/field.class.php
@@ -1260,16 +1260,12 @@ JAVASCRIPT
                     'FKEY' => [
                         'containers' => 'id',
                         'fields' => 'plugin_fields_containers_id',
-                        [
-                            'AND' => [
-                                'containers.itemtypes' => ['LIKE' => "%$itemtype%"]
-                            ]
-                        ]
                     ]
                 ]
             ],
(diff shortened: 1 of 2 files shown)


Mirkk:
Thank you, for me it works now!

inc/field.class.php
```diff
@@ -1260,16 +1260,12 @@ public static function showSingle($itemtype, $searchOption, $massiveaction = fal
...
-                                'containers.itemtypes' => ['LIKE' => "%$itemtype%"]
-                            ]
-                        ]
                     ]
                 ]
             ],
             'WHERE' => [
-                'fields.name' => $cleaned_linkfield
+                'fields.name' => $cleaned_linkfield,
+                'containers.itemtypes' => ['LIKE', "%$itemtype%"]
```
Rom1-B, line 1268: So it should be a "JOIN", not a "LEFT JOIN" (L1258) ?

PR #848, opened 2024-10-17
Fix field Tab Display label

Lainow:

## Description

- It fixes !34825

You can define the tag/label of a field in the fields plugin, but when it is modified it is not updated on the pages and keeps the first name that was defined for it.

### DIFF
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -1139,7 +1139,7 @@ HTML;
         if (isset($itemtypes[$item->getType()])) {
             $tabs_entries = [];
             $container    = new self();
-            foreach ($itemtypes[$item->getType()] as $tab_name) {
+            foreach ($itemtypes[$item->getType()] as $tab_name => $tab_label) {
                 // needs to check if entity of item is in hierachy of $tab_name
                 foreach ($container->find(['is_active' => 1, 'name' => $tab_name]) as $data) {
                     $dataitemtypes = json_decode($data['itemtypes']);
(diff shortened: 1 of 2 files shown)


inc/field.class.php
```diff
@@ -512,7 +512,7 @@ public function getTabNameForItem(CommonGLPI $item, $withtemplate = 0)
         if (!$withtemplate) {
             switch ($item->getType()) {
                 case __CLASS__:
-                    $ong[1] = $this->getTypeName(1);
+                    $ong[1] = "Test";
```
stonebuzz, line 515: :smile:

stonebuzz (changes requested):

inc/container.class.php
```diff
@@ -1139,7 +1139,7 @@ public function getTabNameForItem(CommonGLPI $item, $withtemplate = 0)
         if (isset($itemtypes[$item->getType()])) {
             $tabs_entries = [];
             $container    = new self();
-            foreach ($itemtypes[$item->getType()] as $tab_name => $tab_label) {
+            foreach ($itemtypes[$item->getType()] as $tab_name) {
```
Rom1-B, line 1142: It seems that `$itemtypes[$item->getType()]` can return either a string or an array. The previous code handled the array case, while this one only handles the string case. It needs to work with both.

Lainow: tab label is no longer used as I now look for the value in the base directly

Rom1-B (approved):

stonebuzz (approved):

PR #875, opened 2024-12-19
Fix: API issue preventing multi-field updates

MyvTsv:
## Checklist before requesting a review

*Please delete options that are not relevant.*

- [x] I have performed a self-review of my code.

## Description

- It fixes !35540
- It fixes the issue that prevented modifying a field via the API when a field and a new tab were added to the same object.

PR #892, opened 2025-01-06
Fix PR 875

MyvTsv:

## Description

- It fixes !35784 and https://github.com/pluginsGLPI/fields/pull/875
- Fixes the problem that did not save data when adding an object from a field of the plugin field


trasher:
Replaced by https://github.com/pluginsGLPI/fields/pull/893

PR #910, opened 2025-01-21
Fix api container tab update

### LINKED ISSUE #835: API update does not work with multiple blocks
Hi,

Everything worked fine when I had a single block (id 1) with several fields (76666 -> 76671).

I then created a new block (id 2) with fields 76672 -> 76673.

When I want to update my fields in the 1st block, API says True (OK), but it doesn't actually do anything, whereas it did before.

But on a new block, it works.

If I make block (id 2) inactive, block 1 works again.

The problem occurs as soon as there is more than one block.

GLPI: 10.0.16
Fields: 1.21.13

stonebuzz:

## Description

- It fixes #835

From `API`, plugin need to check from `$item->input` to get related container instead of check `$_REQUEST` whioch is not relevant from `API` process

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [UNRELAESE]
 
+- Fix container update from `API`
+
 ## [1.21.18] - 2024-01-16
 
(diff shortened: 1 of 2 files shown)


stonebuzz:
Ok from customer

PR #857, opened 2024-11-18
fix: datatype number for search

MyvTsv:
- [x] I have performed a self-review of my code.
- [x] I have added tests (when available) that prove my fix is effective or that my feature works.

## Description

- It fixes !35167
- Change the decimal type of number fields to display the decimal part of the decimal numbers in search options

## Screenshots (if appropriate):
Capture d’écran du 2024-11-18 14-11-57

PR #926, opened 2025-03-05
Fix number search option

Lainow:

## Description

- It fixes !36664
Fix the ability to search on 'number' fields 

**Before :**
image

**After :** 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Force float `datatype` of `numeric` fields for fix the search option
 - Fix container update from other context (like plugins)
 
 ## [1.21.19] - 2025-02-03
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-- Force float `datatype` of `numeric` fields for fix the search option
+- Fix `numeric` field search
 - Fix container update from other context (like plugins)
 
 ## [1.21.19] - 2025-02-03
--- a/hook.php
+++ b/hook.php
@@ -364,7 +364,7 @@ function plugin_fields_addWhere($link, $nott, $itemtype, $ID, $val, $searchtype)
             if ($nott) {
                 $link = $link . ' NOT ';
             }
-            return $link . 'CAST(' . $DB->quoteName("$table" . '_' . "$field") . '.' . $DB->quoteName($field) . ' AS DECIMAL(10,2))' . $operator . ' ' . $DB->quoteValue($val) ;
+            return $link . 'CAST(' . $DB->quoteName("$table" . '_' . "$field") . '.' . $DB->quoteName($field) . ' AS DECIMAL(10,9))' . $operator . ' ' . $DB->quoteValue($val) ;
         } else {
             // if 'number' field with name is found with <= or >= or < or > search
             // update WHERE clause with the correct operator
--- a/hook.php
+++ b/hook.php
@@ -341,11 +341,36 @@ function plugin_fields_addWhere($link, $nott, $itemtype, $ID, $val, $searchtype)
     /** @var \DBmysql $DB */
     global $DB;
 
-    $searchopt = &Search::getOptions($itemtype);
-    $table     = $searchopt[$ID]['table'];
-    $field     = $searchopt[$ID]['field'];
+    $searchopt    = &Search::getOptions($itemtype);
+    $table        = $searchopt[$ID]['table'];
+    $field        = $searchopt[$ID]['field'];
+    $pfields_type = $searchopt[$ID]['pfields_type'];
 
     $field_field = new PluginFieldsField();
+
+    if (
+        $field_field->getFromDBByCrit(
+            [
+                'name'     => $field
+            ],
+        )
+        && $pfields_type == 'number'
+    ) {
+        // if 'number' field with name is found with searchtype 'equals' or 'notequals'
+        // update WHERE clause with `$table_$field.$field` because without `$table_$field.id` is used
+        if ($searchtype == 'equals' || $searchtype == 'notequals') {
+            $operator = ($searchtype == 'equals') ? '=' : '!=';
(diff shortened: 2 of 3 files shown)


stonebuzz:
seems a revert of 

https://github.com/pluginsGLPI/fields/pull/857

Can you check whether the issue raised by this PR (857) is actually fixed with it?

Lainow:
> seems a revert of
> 
> #857
> 
> Can you check whether the issue raised by this PR (857) is actually fixed with it?

Yes, the problem has been solved

stonebuzz (approved):

stonebuzz:
Can you validate with customer ?

hook.php
```diff
@@ -341,12 +341,37 @@ function plugin_fields_addWhere($link, $nott, $itemtype, $ID, $val, $searchtype)
     /** @var \DBmysql $DB */
     global $DB;
 
-    $searchopt = &Search::getOptions($itemtype);
-    $table     = $searchopt[$ID]['table'];
-    $field     = $searchopt[$ID]['field'];
+    $searchopt    = &Search::getOptions($itemtype);
+    $table        = $searchopt[$ID]['table'];
+    $field        = $searchopt[$ID]['field'];
+    $pfields_type = $searchopt[$ID]['pfields_type'];
```
stonebuzz, line 347:
```suggestion
    $pfields_type = $searchopt[$ID]['pfields_type'] ?? '';
```
`pfields_type` is added byplugin but is not a regular `key` from glpi core

inc/container.class.php
```diff
@@ -1962,7 +1962,9 @@ public static function getAddSearchOptions($itemtype, $containers_id = false)
                     $opt[$i]['datatype'] = 'text';
                     break;
                 case 'number':
-                    $opt[$i]['datatype'] = 'float';
+                    // change datatype to string because decimal
```
stonebuzz:
```suggestion
                    // change datatype to string because decimal not get `is` / `is not` operator
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Force float `datatype` of `numeric` fields for fix the search option
```
stonebuzz, line 12:
```suggestion
- Fix `numeric` field search
```

stonebuzz (changes requested):

hook.php
```diff
@@ -341,12 +341,37 @@ function plugin_fields_addWhere($link, $nott, $itemtype, $ID, $val, $searchtype)
...
 
     $field_field = new PluginFieldsField();
 
+    if (
+        $field_field->getFromDBByCrit(
+            [
+                'name'     => $field
+            ],
+        )
+        && $pfields_type == 'number'
```
stonebuzz:
```suggestion
        $field_field->getFromDBByCrit(
            [
                'name'     => $field,
                'type' => 'number'
            ],
        )
        && $pfields_type == 'number'
```

just an additional control

stonebuzz (changes requested):

inc/container.class.php
```diff
@@ -1962,7 +1962,9 @@ public static function getAddSearchOptions($itemtype, $containers_id = false)
                     $opt[$i]['datatype'] = 'text';
                     break;
                 case 'number':
-                    $opt[$i]['datatype'] = 'decimal';
+                    // change datatype to string to get `is` / `is not` operator
+                    $opt[$i]['datatype'] = 'string';
+                    $opt[$i]['searchtype'] = ['contains', 'notcontains', 'equals', 'notequals'];
```
Rom1-B, line 1967: Just adding `searchtype` wouldn't be enough?
```suggestion
                    $opt[$i]['datatype'] = 'decimal';
                    $opt[$i]['searchtype'] = ['contains', 'notcontains', 'equals', 'notequals'];
```

Rom1-B (changes requested):

inc/container.class.php
stonebuzz: decimal type have a different behavior from Search.class to handle missing `.`

                if (isset($searchopt[$ID]["datatype"]) && ($searchopt[$ID]["datatype"] === 'decimal')) {
                    $matches = [];
                    if (preg_match('/^(\d+.?\d?)/', $val, $matches)) {
                        $val = $matches[1];
                        if (!str_contains($val, '.')) {
                            $val .= '.';
                        }
                    }
                }

Sa we cannot handle all type of search; knowing that this field is in text format, and can be used to save numeric or decimal fields (a nightmare)

stonebuzz: in my opinion, with your change, not all of the following conditions will be functional

is 15 (match only 15 not 15.50 or other)
is not 15 (match other expect 15)
contain 15 (match 15.10, 10.15, 15)
not contain 15 (match other than 15.10, 10.15, 15)
>= 15 (match 15.05 , 15, 16)
<= 15 (match 10, ,10.50, 15)

hook.php
```diff
@@ -341,12 +341,42 @@ function plugin_fields_addWhere($link, $nott, $itemtype, $ID, $val, $searchtype)
...
+    ) {
+        // if 'number' field with name is found with searchtype 'equals' or 'notequals'
+        // update WHERE clause with `$table_$field.$field` because without `$table_$field.id` is used
+        if ($searchtype == 'equals' || $searchtype == 'notequals') {
+            $operator = ($searchtype == 'equals') ? '=' : '!=';
+            if ($nott) {
+                $link = $link . ' NOT ';
+            }
+            return $link . 'CAST(' . $DB->quoteName("$table" . '_' . "$field") . '.' . $DB->quoteName($field) . ' AS DECIMAL(10,2))' . $operator . ' ' . $DB->quoteValue($val) ;
```
stonebuzz, line 367:
```suggestion
            return $link . 'CAST(' . $DB->quoteName("$table" . '_' . "$field") . '.' . $DB->quoteName($field) . ' AS DECIMAL(10,9))' . $operator . ' ' . $DB->quoteValue($val) ;
```

In case when user have saved decimal like 10.2568 (with more than two digit after comma)

stonebuzz (changes requested):

stonebuzz (approved):

Rom1-B (approved):

stonebuzz:
@Lainow can you rebase ?

Lainow:
Ok for customer

PR #932, opened 2025-03-20
1.21.20

stonebuzz:
### Fixed

- Fix `numeric` field search
- Fix containers migration while adding `is_recursive` field
- Fix container update from other context (like plugins)
- Fix "not equals" search operator for dropdown `multiple`
- Fix container data (`entities_id`) insert from `dom` type

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -100,7 +100,7 @@ Il existe un [script de migration](https://github.com/pluginsGLPI/customfields/b
    <versions>
       <version>
          <num>1.21.20</num>
-         <compatibility>~10.0.0</compatibility>
+         <compatibility>~10.0.11</compatibility>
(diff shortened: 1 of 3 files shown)


plugin.xml
```diff
@@ -98,6 +98,11 @@ Il existe un [script de migration](https://github.com/pluginsGLPI/customfields/b
       <author>Olivier Moron</author>
    </authors>
    <versions>
+      <version>
+         <num>1.21.20</num>
+         <compatibility>~10.0.0</compatibility>
```
Rom1-B, line 103:
```suggestion
         <compatibility>~10.0.11</compatibility>
```

Rom1-B (changes requested):

Rom1-B (approved):

PR #937, opened 2025-03-26
Fix - Validation for mandatory multiple dropdown

RomainLvr:

## Description

- It fixes !37010

If a required multiple dropdown field is present in our ticket form, it is currently possible to create a ticket without selecting any values.

The issue comes from the validation function, which only checks if a required field is empty but does not properly handle multiple-selection fields. For a standard dropdown, the function considers a value of 0 as empty. However, for a multiple dropdown, an unselected field results in an empty array, which is not detected as invalid.

A specific condition must be added to check if a required multiple-selection field is empty by detecting empty arrays during validation.

image

### DIFF
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -1521,9 +1521,7 @@ HTML;
             if (
                 $field['mandatory'] == 1
                 && (
-                    $value === null
-                    || $value === ''
-                    || (is_array($value) && empty($value))
+                    empty($value)
                     || (($field['type'] === 'dropdown' || preg_match('/^dropdown-.+/i', $field['type'])) && $value == 0)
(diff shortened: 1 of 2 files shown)


inc/container.class.php
```diff
@@ -1523,6 +1523,7 @@ public static function validateValues($data, $itemtype, $massiveaction)
                 && (
                     $value === null
                     || $value === ''
+                    || (is_array($value) && empty($value))
```
Rom1-B, line 1526:
```suggestion
                    empty($value)
```

Rom1-B (changes requested):

stonebuzz (approved):

Issue #834, opened 2024-09-23
Fields plugin not compatible with GLPI entities


blippercop:
Using GLPI with multiple entities is not working for the fields plugin.

Company > Unit 1
Company > Unit 2
Company > Unit 3

When the field is created on "Company" level with inheritance, all units can see the fields. However when one unit requires a second fields (specifically for the unit with Insertion in the form (before save button)) it is not working because of an error:

_You cannot add several blocks with type 'Insertion in the form' on same object_

But when you create the fields for each unit (instead of Company level with inheritance) you get the same error. This should be possible as it is still "one" form but for each unit.

CorentinS6:
Hello,
+1
see #751 #789 #744

Issue #789, opened 2024-05-21
Impossible to use same block type in several entities


tsadock:
Hi,

It is impossible to create two blocks, in two different entities with the same item type.
It raises this message : You cannot add several blocks with type 'Insertion in the form' on same object

Is it possible to add "entity" check when adding a new block ?

Kind regards

CorentinS6:
Hi,
see #744 ?

Issue #744, opened 2023-12-18
You cannot add several blocks with type 'Insertion in the form' on same object (Different entity)


berilslv:
Hello,

Thanks for developing this great plugin, but I'm facing issue:

I'm trying to create another additional field block in dropdown->locations section.
I have already created additional fields in location for Entity A. Now I need additional fields in location section for entity B. But I'm getting error bellow:
 You cannot add several blocks with type 'Insertion in the form' on same object

Is there any workaround how can I create custom additional fields in location section for each entity?

Thanks!
p.s. running on 10.0.10 , ubuntu 22.04

CorentinS6:
same problem here.
Waiting a clean solution, it seems to work with a dirty sql query : set type to "dom" after bloc creation (table : glpi_plugin_fields_containers)

stonebuzz:
The plugin needs to generate PHP files to work

Modifying the values in the database is clearly not a good idea.

it's clearly intended to prevent modification after creation

best regards

CorentinS6:
ok @stonebuzz, but the issue is not solved...

stonebuzz:
To complete my remarks, 
The "fields" plugin is not currently designed to think in terms of entities.

The "Insert before saved button" type can only be applied once per object (I won't go into the technical details, but it's a question of container identifiers).

However, the "Add a tab" type can be used several times.
In this case, you would simply need to add conditions to hide the tabs depending on the entity.

![image](https://github.com/pluginsGLPI/fields/assets/7335054/7051641b-16f6-40a7-9280-1a96a434a871)

Best regards

PR #941, opened 2025-04-04
[Feature] Allow multiple "Insertion in the form" (dom) blocks per entity

JeremieMercier:
This PR provides the ability to define multiple "Insertion in the form" blocks for the same item based on its entity

- Created a new method findContainers() (based on findContainer()) that returns all 'dom' containers for an item based on its entity (with parent entity handling via getAncestorsOf()).
- Adapted hooks (pre_item_add, pre_item_update, post_item_add, post_item_update) to manage multiple containers using the _plugin_fields_data_multi array.
- Updated the populateData() function to extract input values by stripping the prefix, ensuring that data is saved into the correct columns of the injection table.
- Modified the container.form.php file to "clean" the form data (by removing the prefix) before calling updateFieldsValues(), thereby enabling the saving of domtab containers.

Please test these modifications and verify that everything works as expected for all object types supported by the Fields plugin. Note that my tests were successful, although I focused exclusively on 'Ticket' type objects for my use case.

- [x] This change requires a documentation update.

Issues : #744, #789, #834


stonebuzz:
Hi @JeremieMercier 

I am questioning the plugin's ability to reconcile the correct container during API calls, an issue we have encountered in the past. For example, when updating a ticket and including the "fields" in the payload, could you verify this point?

JeremieMercier:
I retrieve the block row ID like this:
image

Then, I update the field using the following method:
image

So I haven’t encountered any issue with the API — unless I misunderstood how you're updating the fields via the API?

stonebuzz:
It is also possible to update fields from the fields plugin during ticket update.

`PUT {{URL_GLPI}} /Ticket/5`

JeremieMercier:
Yes, it works!

Just include the block ID in the field name, like:
plugin_fields_15_bloconechamponefield

image
image

JeremieMercier:
Hello, do you have time to review the PR ?

inc/container.class.php
```diff
@@ -1604,6 +1588,83 @@ public static function findContainer($itemtype, $type = 'tab', $subtype = '')
...
+                $glpiActiveEntities = $_SESSION['glpiactiveentities'] ?? 0;
+
+                $entityRestriction = getEntitiesRestrictCriteria('', '', $glpiActiveEntities, true, true);
+
+                if (count($entityIds) === 0) {
+                    $entityIds = [$entityId];
+                }
+                $entityIdList = implode(",", $entityIds);
+
+                $sql = "SELECT id FROM glpi_plugin_fields_containers
```
trasher, line 1614: You must use [GLPI DBIterator](https://glpi-developer-documentation.readthedocs.io/en/develop/devapi/database/dbiterator.html) instead of direct queries.

JeremieMercier:
A syntax error, it was slipped in after applying the suggestion 😅

JeremieMercier:
Are we good this time ? 😅🤞

JeremieMercier:
Hi, what's the next step ?

stonebuzz:
Hi @JeremieMercier 

The feature you are proposing is substantial and introduces significant changes. To ensure its smooth integration and functionality, we must be extremely vigilant. This means we need to "lock down" our testing to guarantee that everything works as expected.

It is important to acknowledge that the plugin is already complex to maintain, especially with the new features introduced in GLPI. Given this complexity, we must approach any additional features with caution to ensure they do not further complicate maintenance and support.

To ensure the proper functioning of this PR, here is a concise list of unit tests that should be added:

### Unit Tests for `findContainers`

1. **Retrieve Containers for Specific Itemtype and Entity**
   - Verify that `findContainers` returns the correct containers for a given itemtype and entity.

2. **Retrieve Containers with Subtype**
   - Verify that `findContainers` returns the correct containers when a subtype is specified.

3. **Retrieve Containers with Recursive Entity**
   - Verify that `findContainers` considers recursive entities.

### Unit Tests for Hooks

4. **`preItemAdd` with Multiple Containers**
   - Verify that the `preItemAdd` hook correctly handles data for multiple containers.

5. **`postItemAdd` with Multiple Containers**
   - Verify that the `postItemAdd` hook correctly updates field values for multiple containers.

6. **`preItemUpdate` with Multiple Containers**
   - Verify that the `preItemUpdate` hook correctly handles data for multiple containers during item update.

### Unit Tests for `populateData`

7. **`populateData` with Prefix**
   - Verify that `populateData` correctly extracts field values by removing the prefix.

8. **`populateData` with Multiple Selection Fields**
   - Verify that `populateData` correctly handles fields allowing multiple selections.

### Unit Tests for `showForTab`

9. **`showForTab` with Multiple Containers**
   - Verify that `showForTab` correctly displays containers for a given item.

10. **`showForTab` with Insufficient Rights**
    - Verify that `showForTab` returns nothing if the user does not have sufficient rights on the containers.

### Unit Tests for `API`

11. **Ticket input update**

12. **Container update**

JeremieMercier:
Hi @stonebuzz Thanks for the message

I fully understand the concerns and the importance of thorough testing.
I'm currently working on implementing the required tests and will make sure all the listed cases are properly covered.

While working on the tests already partially in place, I’ve been able to identify and fix some issues especially related to "dropdown" field handling and container restriction management.

I’ll keep updating here as I progress.

akstis-typer:
Hi, @JeremieMercier.
What tool you are using to send HTTP requests?
I'm newby in WEB development so I'm not entirely shure what tool is that.
Thank you :)

JeremieMercier:
I’ve added all the requested unit tests (except the API ones):

- findContainers : basic cases, sub-types, entity recursion
- Hooks: preItem, postItemAdd, preItemUpdate with multiple containers
- populateData: prefix handling and multi-selection fields
- showForTab: displaying multiple containers and checking rights

Please note I’m not a testing expert, so the tests may not follow every usual best practice. Feel free to share any feedback so I can improve.

The API tests still need to be written by whoever is willing to tackle them.

CorentinS6:
And code generated by AI ?

JeremieMercier:
> And code generated by AI ?

Hello, All the logic is 100% mine, however for the tests, it's another story, I used AI to learn and understand the use of the tests in addition to videos like this one from Grafikart, but as I said, I'm not an expert..., the tests can also be redone by a more competent person if necessary knowing that basically, there were no tests set up on the plugin.

CorentinS6:
@stonebuzz , is this work good enough for the merge now?

Rom1-B (changes requested):

front/container.form.php
```diff
@@ -57,6 +57,17 @@
...
+        foreach ($_REQUEST as $key => $value) {
+            // if key starts with plugin_fields_<containerID>_ remove the prefix
+            if (strpos($key, "plugin_fields_{$containerID}_") === 0) {
+                $new_key = substr($key, strlen("plugin_fields_{$containerID}_"));
+                $data[$new_key] = $value;
+            } else {
+                $data[$key] = $value;
+            }
+        }
         $container->updateFieldsValues($_REQUEST, $_REQUEST['itemtype'], false);
```
Rom1-B: `$data` is computed but never used?!

inc/container.class.php
```diff
@@ -1774,68 +1845,57 @@ public static function preItemUpdate(CommonDBTM $item)
...
+            if (false !== ($data = self::populateData($c_id, $item))) {
+                if (!self::validateValues($data, $item->getType(), isset($_REQUEST['massiveaction']))) {
+                    // if validation fails, we need to remove the data from the item input
+                    $item->input = [];
 
-            return true;
+                    return false;
+                }
+                $all_data[] = $data;
+            }
```
Rom1-B, line 2040: When an item has several applicable containers, a validation failure on any single container wipes `$item->input` and returns immediately, discarding the data already accumulated in `$all_data` for the other, valid containers. In the single-container world this was the existing behaviour (nothing else to lose), but with multiple containers per item it means one invalid field in container B now also blocks saving valid data in container A. Please confirm whether this all-or-nothing semantics is intentional; if not, previously-validated containers' data should still make it into `_plugin_fields_data_multi`.

inc/field.class.php
```diff
@@ -863,8 +863,13 @@ public static function showDomContainer($id, $item, $type = 'dom', $subtype = ''
             $fields = [];
         }
 
+        $fieldTypeName = '_plugin_fields_type_' . $id;
+        $fieldSubTypeName = '_plugin_fields_subtype_' . $id;
+
         echo Html::hidden('_plugin_fields_type', ['value' => $type]);
         echo Html::hidden('_plugin_fields_subtype', ['value' => $subtype]);
+        echo Html::hidden($fieldTypeName, ['value' => $type]);
+        echo Html::hidden($fieldSubTypeName, ['value' => $subtype]);
```
Rom1-B: `_plugin_fields_type_{id}` / `_plugin_fields_subtype_{id}` are added alongside the existing non-suffixed hidden fields, but no other code (PHP, Twig, or JS) in the plugin reads them back. Either wire them up (presumably needed to disambiguate type/subtype per container once several `dom` containers coexist on one page) or drop them.

composer.lock
Rom1-B, line 1: The diff downgrades `sebastian/diff` from `6.0.2` (PHP `>=8.2`, dev-requires `phpunit/phpunit ^11.0`) to `4.0.6` (PHP `>=7.3`, dev-requires `phpunit/phpunit ^9.3`), alongside an unrelated `phpstan/phpstan` patch bump. This looks like stale lockfile drift from an out-of-sync branch rather than an intentional change for this feature. Suggest regenerating the lockfile with a clean `composer update` against current `composer.json` before merge.

tests/PluginFieldsContainerTest.php
```diff
@@ -0,0 +1,753 @@
...
+        $this->addFieldToContainer($containerId2, 'field2');
+
+        // add a ticket
+        $ticket = $this->addTicket([
+            'name'                                      => 'Test ticket postItemAdd',
+            'content'                                   => 'test content',
+            'entities_id'                               => 0,
+        ]);
+
+        // 3. Injecter les données comme si elles venaient du formulaire
```
Rom1-B: Should be in English per the project's coding conventions.

Rom1-B (changes requested):

front/container.form.php
JeremieMercier: yes that was a true bug : the prefix-stripping loop built $data but the call still passed $_REQUEST, so prefixed keys were never matched by updateFieldsValues() (which expects unprefixed keys). fixed the cleaned $data is now passed.

inc/container.class.php
JeremieMercier: Yes the all-or-nothing semantics is intentional all applicable blocks are rendered and submitted within a single form, so a validation failure anywhere blocks the whole submission the user gets the error messages from validateValues(), fixes the input and resubmits. Partially saving would be worse : the item would be created with container A's data while container B's data is silently dropped. This also matches the historical single-container behaviour, simply extended to several containers

inc/field.class.php
JeremieMercier: You're right they were dead code an early attempt at per-container disambiguation that became unnecessary once preItem() resolved applicable containers itself via findContainers(). They have been dropped only the non-suffixed _plugin_fields_type / _plugin_fields_subtype fields remain

JeremieMercier:
I originally opened this PR because it was an internal feature request for how we use the GLPI plugin. In the end, we managed without it, so I kind of let it drop especially with the commit lag and conflicts to deal with.

Anyway, I've decided to pick it back up: even though we don't need the feature ourselves anymore, it could still be useful to others since it's quite heavily requested ^^

inc/container.class.php
```diff
@@ -1891,87 +1959,76 @@ public static function preItemUpdate(CommonDBTM $item)
...
+                if (!self::validateValues($data, $item->getType(), isset($_REQUEST['massiveaction']))) {
+                    // if validation fails, we need to remove the data from the item input
+                    $item->input = [];
 
-                return false;
+                    return false;
+                }
+                $all_data[] = $data;
+                continue;
             }
```
Rom1-B, line 2022: `preItem()` loops over every `dom` container applicable to the item. Before this PR there was only ever one container per itemtype, so returning `false` on the first validation failure was harmless. Now that several containers can apply to the same item, if container A is valid and container B is not, the loop still wipes `$item->input` and returns `false` as soon as B fails ; discarding A's already-validated data too. One invalid container therefore blocks saving every other valid container on the same item.

@JeremieMercier is this all-or-nothing behavior intentional, or should valid containers still be saved when another one fails validation? If the latter, previously-validated containers' data should still be pushed into `_plugin_fields_data_multi` instead of being wiped.

JeremieMercier: Yes the all-or-nothing semantics is intentional all applicable blocks are rendered and submitted within a single form, so a validation failure anywhere blocks the whole submission the user gets the error messages from validateValues(), fixes the input and resubmits. Partially saving would be worse : the item would be created with container A's data while container B's data is silently dropped. This also matches the historical single-container behaviour, simply extended to several containers

Rom1-B (changes requested):
Several errors in the CI

JeremieMercier:
hi @Rom1-B, and now ?

PR #942, opened 2025-04-10
Fixed issue with hiding conditions not hiding containers in simple ticket view

### LINKED ISSUE #847: Hiding fields depending on category does not work
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [X] I have searched the existing issues

### GLPI Version

10.0.15

### Plugin version

1.21.15

### Bug description

When I setup a block to be hidden depending on ticket category, block is not hidden in simplified interface. Haven't tried in standard interface since it does not meet our requirements.

Block configuration:

Block config 1

Block fields:

Block fields

Block hiding condition:

Block condition
Block condition 2

When a user begins to create a ticket, block is shown (which is expected to happen)

Ticket creation before setting category

But when the user sets the category by which the block has to be hidden, some erratic movement happens and the field is still shown with broken CSS, a.k.a. moved rightwards and wider.

Ticket creation after setting category

I've tried setting the condition to the ticket's title and this works correctly, block is successfully hidden and works like a charm.

### Relevant log output

_No response_

### Page URL

_No response_

### Steps To reproduce

1. Create a block and set it's type to "Insertion in the form (before save button)".
2. Configure some fields in the block.
3. Set a hiding conditional rule that depends on ticket's category.
4. Test a ticket creation by setting the category specified on step `3` and block is still present.

### Your GLPI setup information

_No response_

### Anything else?

_No response_

akstis-typer:
- It fixes #847

### DIFF
--- a/ajax/container.php
+++ b/ajax/container.php
@@ -56,12 +56,19 @@ if (isset($_GET['action']) && $_GET['action'] === 'get_fields_html') {
 
     $display_condition = new PluginFieldsContainerDisplayCondition();
     if ($display_condition->computeDisplayContainer($item, $containers_id)) {
+        $field_options = [
+            'label_class' => 'col-lg-3',
+            'input_class' => 'col-lg-9',
+        ];
+        echo "<div class='offset-md-1 col-md-8 col-xxl-6'>";
         PluginFieldsField::showDomContainer(
             $containers_id,
             $item,
             $type,
             $subtype,
+            $field_options,
         );
+        echo "</div>";
     } else {
         echo '';
     }
--- a/inc/field.class.php
+++ b/inc/field.class.php
@@ -887,6 +887,12 @@ class PluginFieldsField extends CommonDBChild
     {
         $item = $params['item'];
 
+        if ($item->fields['type'] == "") {
+            $item->fields['type'] = $params['options']['type'];
+        }
+        if ($item->fields['itilcategories_id'] == "") {
+            $item->fields['itilcategories_id'] = $params['options']['itilcategories_id'];
+        }
         $functions = array_column(debug_backtrace(), 'function');
         $subtype   = isset($_SESSION['glpi_tabs'][strtolower($item::getType())]) ? $_SESSION['glpi_tabs'][strtolower($item::getType())] : '';
         $type      = substr($subtype, -strlen('$main')) === '$main'
@@ -956,7 +962,7 @@ class PluginFieldsField extends CommonDBChild
         }
 
         $html_id = 'plugin_fields_container_' . mt_rand();
-        if (strpos($current_url, 'helpdesk.public.php') !== false) {
+        if (strpos($current_url, 'helpdesk.public.php') !== false || strpos($current_url, 'tracking.injector.php') !== false) {
             echo "<div id='{$html_id}' class='card-body row mx-0' style='border-top:0'>";
             echo "<div class='offset-md-1 col-md-8 col-xxl-6'>";
             $field_options = [
@@ -976,7 +982,7 @@ class PluginFieldsField extends CommonDBChild
                 $field_options ?? [],
             );
         }
-        if (strpos($current_url, 'helpdesk.public.php') !== false) {
+        if (strpos($current_url, 'helpdesk.public.php') !== false || strpos($current_url, 'tracking.injector.php') !== false) {
             echo '</div>';
         }
         echo '</div>';


DaviidMM:
Just to  clarify, in my installation (new, all of it with the last versions) I had to place the code in `showForTab` funcion below the line `$display_condition = new PluginFieldsContainerDisplayCondition();`:

```php
...

// @akstis-typer code starts here ------
if($item->fields['type'] == "")
{
    $item->fields['type'] = $params['options']['type'];
}
if($item->fields['itilcategories_id'] == "")
{
    $item->fields['itilcategories_id'] = $params['options']['itilcategories_id'];
}
// @akstis-typer code ends here ------

if ($display_condition->computeDisplayContainer($item, $c_id)) {
...
```

DaviidMM:
I've been testing this code and works correctly except for one more issue: When a custom field is shown and you edit another field (e.g. title), the custom field's alignment breaks again. My solution was to add `echo "<div class='offset-md-1 col-md-8 col-xxl-6'>";` into the `ajax/container.php` file, inside the `if ($display_condition->computeDisplayContainer($item, $containers_id)) {` condition:

```php
    if ($display_condition->computeDisplayContainer($item, $containers_id)) {
        $field_options = [
            'label_class' => 'col-lg-3',
            'input_class' => 'col-lg-9',
        ];
        echo "<div class='offset-md-1 col-md-8 col-xxl-6'>";
        PluginFieldsField::showDomContainer(
            $containers_id,
            $item,
            $type,
            $subtype,
            $field_options
        );
        echo "</div>";
    } else {
        echo '';
    }
```

stonebuzz:
Hi @akstis-typer 

can you rebase ?

akstis-typer:
> Hi @akstis-typer
> 
> can you rebase ?

Hi, you mean I need to add last changes of original repository? Or something else?
I'm not so good at git so I didn't get it :0

stonebuzz:
```
git fetch upstream (github repo)
git pull --rebase upstream main
git push --force 
```

That should be sufficient.

akstis-typer:
I think I did it right. The next step is unit tests - how do I make them? Or can I run yours?

composer.json
```diff
@@ -5,7 +5,6 @@
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^3.75",
-        "friendsoftwig/twigcs": "^6.1",
```
trasher, line 8: I do not think it should be removed.

trasher (commented):
You PR has lot of unrelated changes (new lines added, spacing added/removed) that must be reverted.

composer.json
Rom1-B: I agree: we have to keep it.

akstis-typer:
I've done poorly last time and accidentally revert changes in twig files. Now I reverted all back how it should be and removed spaces and new lines I added.

templates/fields.html.twig
Rom1-B, line 1: This file has no reason to be modified

templates/status_overrides.html.twig
Rom1-B, line 1: This file has no reason to be modified

twig_cs.dist.php
Rom1-B, line 1: This file must keep its original name

Rom1-B (changes requested):

akstis-typer:
Finally I got it

leeopereira:
=

> I've been testing this code and works correctly except for one more issue: When a custom field is shown and you edit another field (e.g. title), the custom field's alignment breaks again. My solution was to add `echo "<div class='offset-md-1 col-md-8 col-xxl-6'>";` into the `ajax/container.php` file, inside the `if ($display_condition->computeDisplayContainer($item, $containers_id)) {` condition:
> 
> ```
> $display_condition = new PluginFieldsContainerDisplayCondition();
>     if ($display_condition->computeDisplayContainer($item, $containers_id)) {
>         $field_options = [
>             'label_class' => 'col-lg-3',
>             'input_class' => 'col-lg-9',
>         ];
>         echo "<div class='offset-md-1 col-md-8 col-xxl-6'>";
>         PluginFieldsField::showDomContainer(
>             $containers_id,
>             $item,
>             $type,
>             $subtype,
>             $field_options
>         );
>         echo "</div>";
>     } else {
>         echo '';
>     }
> ```
> 
> After the `showDomContainer` method i close the created div with `echo "</div>";`



I tested it and in my environment it only worked correctly this way.

Thank you very much everyone!

leeopereira:
Great, congratulations!
You updated the plugin, but when I applied the update, not only did it not fix this problem, but the fix from the colleague above no longer works.
Great!

leeopereira:
> Hi, @leeopereira, my fix doesn't work on the nightly version of the plugin?

Oi @akstis-typer , Unfortunately not, I had to restore a backup of the previous version of the files. When I updated, the deleted files were replaced with the originals from the new version, and the problem returned. The hide condition stopped working, and the custom fields became misaligned. :/

akstis-typer:
@leeopereira , this happened because this changes are not in the main branch of fields plugin repo, so when you update using glpi - they getting overriden by update that developers made.
But, if you pull changes from this branch and merge it with updated branch - it'll still work :)
I may update my branch to the nightly one, and you'll be able to pull it without problem.

leeopereira:
@akstis-typer After updating, I tried editing the files and inserting the corrections here in the topic, but it still didn't work. I don't know if they changed something in the code that impacted the correction. But I just returned the two old edited files and it worked.

ajax/container.php
```diff
@@ -56,12 +56,19 @@
...
+        ];
+        echo "<div class='offset-md-1 col-md-8 col-xxl-6'>";
         PluginFieldsField::showDomContainer(
             $containers_id,
             $item,
             $type,
             $subtype,
+            $field_options,
         );
+        echo "</div>";
```
Rom1-B, line 71: This wrapper is now always applied on AJAX refresh, but `showForTab` (`inc/field.class.php:965-974`) only applies it for `helpdesk.public.php`/`tracking.injector.php`. Normal admin/agent forms load with no wrapper, then get re-wrapped as soon as any field changes ; fields shift after the first edit. Matches misalignment reports on the PR thread (e.g. leeopereira, 2025-08-28).

Should this only apply when `$type`/`$subtype` matches the public/injector context, like `showForTab`?

inc/field.class.php
```diff
@@ -887,6 +887,12 @@ public static function showForTab($params)
     {
         $item = $params['item'];
 
+        if ($item->fields['type'] == "") {
+            $item->fields['type'] = $params['options']['type'];
+        }
+        if ($item->fields['itilcategories_id'] == "") {
+            $item->fields['itilcategories_id'] = $params['options']['itilcategories_id'];
+        }
```
Rom1-B, line 895: `post_item_form` is a global hook, fired for all 70+ itemtypes, not just `Ticket`. Most have no `type`/`itilcategories_id` column (even `Change`/`Problem` lack `type`), so this raises "Undefined array key" warnings on nearly every item's edit form.

```suggestion
        if ($item instanceof Ticket) {
            $item->fields['type'] = $item->fields['type'] ?: ($params['options']['type'] ?? '');
        }
        if ($item instanceof CommonITILObject) {
            $item->fields['itilcategories_id'] = $item->fields['itilcategories_id'] ?: ($params['options']['itilcategories_id'] ?? '');
        }
```

Can you test with a non-ITIL item (e.g. `Computer`)? No test covers this method.

Rom1-B (changes requested):

PR #953, opened 2025-05-20
Fic(Condition): Fix condition check logic for dropdown field values

### LINKED ISSUE #947: Plugin does not respect category conditions
### Code of Conduct

- [ ] I agree to follow this project's Code of Conduct

### Is there an existing issue for this?

- [ ] I have searched the existing issues

### GLPI Version

10.0.16

### Plugin version

1.21.21

### Bug description

Hello to all

When restricting a block with a condition, it is not respected according to those conditions

Here are the steps to follow to replicate the case

### Relevant log output

```shell

```

### Page URL

_No response_

### Steps To reproduce

1. Select the condition to hide the block. In my case, there are two conditions
Image

2. It should hide the field when the category is different from “PEI > Pendrive > Provisioning Request” or from “PEI > Cellular Equipment > Roaming Management”
Image

3. The category is equal to: “PEI > Cellular Equipment > Roaming Management”, and the block does not appear in the ticket.
Image

### Your GLPI setup information

_No response_

### Anything else?

_No response_

stonebuzz:

## Description

- It fixes #947

Prevent duplicate call only from new object

### DIFF
--- a/inc/field.class.php
+++ b/inc/field.class.php
@@ -1033,9 +1033,9 @@ class PluginFieldsField extends CommonDBChild
                         'change',
                         'input, select, textarea',
                         function(evt) {
-                            if (evt.target.name == "itilcategories_id") {
+                            if (evt.target.name == "itilcategories_id" && {$items_id} == 0) {
                                 // Do not refresh tab container when form is reloaded
-                                // to prevent issues diues to duplicated calls
+                                // to prevent issues diues to duplicated calls (when object is new)
                                 return;
                             }
                             if ($(evt.target).closest('#{$html_id}').length > 0) {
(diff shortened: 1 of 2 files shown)


inc/field.class.php
```diff
@@ -1033,9 +1033,9 @@ function () {
                         'change',
                         'input, select, textarea',
                         function(evt) {
-                            if (evt.target.name == "itilcategories_id") {
+                            if (evt.target.name == "itilcategories_id" && {$items_id} == 0) {
```
trasher, line 1036:
```suggestion
                            if (evt.target.name == "itilcategories_id" && $items_id == 0) {
```

stonebuzz: Are you sure ? other php variable are enclosed by `{}`

Rom1-B: I also think it should be enclosed, because it's PHP that generates JS.

trasher: Yes, indeed, my bad

stonebuzz: :beers:

PR #963, opened 2025-06-17
Fix typo

trasher:


stonebuzz:
Can you please revert the changes made to the locales/*.po files?

The GitHub Actions workflow for translations expects updates to be made in these files only when running the extraction process (extract-locales).

see :
https://github.com/glpi-project/plugin-translation-workflows/blob/f6ed6e4b91955d6913a0b471a0934550600c8306/.github/workflows/transifex-sync.yml#L59-L64

PR #980, opened 2025-07-17
fix(SO): search options not working with default values in multi-select

MyvTsv:

## Description

- It fixes !38487

When a multiple drop-down list field, defined via the Fields plugin, has a default value, this is displayed in the interface when a new item is created (itemtype), but is not saved in the database.

This leads to inconsistent behavior: when GLPI filters search for this value, items supposed to have this default value are not found, as it is not actually saved.

image
image

### DIFF
--- a/hook.php
+++ b/hook.php
@@ -376,7 +376,6 @@ function plugin_fields_addWhere($link, $nott, $itemtype, $ID, $val, $searchtype)
             }
         }
     }
-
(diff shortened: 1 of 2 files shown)


stonebuzz (approved):

stonebuzz:
@MyvTsv can you valide with customer?

Rom1-B (changes requested):
Seems OK
Waiting customer validation.

MyvTsv:
It's good for the customer

PR #985, opened 2025-07-23
fix(searchoption): default value not displayed when field is not set

MyvTsv:

## Description

- It fixes !38487
- When creating an object that has fields in a tab created by the field plugin, if the fields have a default value, they are not saved to the database. In this case, I check whether the row is saved to the database. If the value is not present, the default value is displayed.

**Before:**
image

**After:**
image


Rom1-B (changes requested):
Seems OK.
Please have the customer approve it and we'll merge it afterwards.

Rom1-B (approved):

PR #1063, opened 2025-10-29
fix(user): Align the fields of the plugin with the user form

MyvTsv:

## Description

- It fixes !39926
- The fields in the native user form were not correctly aligned with the fields created by the Fields plugin. 

Before fix:
image

Add User: After Fix:
image

Update User: After fix
image

### DIFF
--- a/templates/fields.html.twig
+++ b/templates/fields.html.twig
@@ -42,6 +42,9 @@
         <div class="d-flex flex-row flex-wrap flex-xl-nowrap">
             <div class="row flex-row align-items-start flex-grow-1">
                 <div class="row flex-row">
+                    {% if item.isNewItem() and item is instanceof('User') %}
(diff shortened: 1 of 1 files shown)


templates/fields.html.twig
```diff
@@ -37,7 +37,7 @@
 
 {% if not already_wrapped and not dropdown_item%}
 
-{% set class = item.isNewItem() ? 'col-xxl-12' : 'col-xxl-9' %}
+{% set class = item.isNewItem() or item is instanceof('User') ? 'col-xxl-12' : 'col-xxl-9' %}
```
Rom1-B: For existing users, it is well aligned, but for new users, there is a slight misalignment:

Image

Rom1-B (approved):

PR #1074, opened 2025-11-04
fix $item->input that could not be an array in container.class.php

CorentinS6:
fix if $item->input not an array

## Description

- It fixes # (issue number, if applicable)


stonebuzz:
I agree with @trasher , moreover,

the issue seems to come from the `metademands` plugin instead.

PR #1084, opened 2025-11-12
fix(fields): hide menu without config read rights

MyvTsv:

## Description

- It fixes !40238
If a user profile does not have read access to the GLPI configuration, it should also not have access to the Fields plugin containers.

However, the user could still see the Fields menu in the Setup menu, even though they could not view its contents due to insufficient permissions.

This PR corrects this behavior by hiding the Fields menu when the user does not have the corresponding read permission.

image


stonebuzz (approved):

Rom1-B:
Please rebase

PR #1118, opened 2026-01-20
fix(migration): resolve compile error during plugin update with genericobject asset

MyvTsv:

## Description

- It fixes !41370
- When migrating the Fields plugin, container files are regenerated. However, if the targeted itemtype classes are not loaded, the file is not updated, which poses a problem in the event of a major change.

In GLPI 10, this behavior occurs when a container targets a deleted GenericObject: the CustomAsset is then not migrated to version 11 and the file remains obsolete.

`[2026-01-20 13:57:37] glpi.ALERT:   *** Fatal Compile Error: Declaration of PluginFieldsPlugingenericobjectprojecteurprojecteurtab::addField($fieldname, $type, array $options) must be compatible with PluginFieldsAbstractContainerInstance::addField($fieldname, $type, array $options = [])
  Backtrace :
  ...ericobjectprojecteurprojecteurtab.class.php:169`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Add compatibility with GLPI `CustomAsset`
 
+### Fixed
+
+- Fix migration error caused by unknown itemtype in containers
 
 ## [1.23.2] - 2025-12-22
 


Rom1-B (changes requested):
OK, I'm not approving it just to block the merge and wait for other cases.

PR #1122, opened 2026-01-29
Feat(Core): Implement HL API

stonebuzz:

## Description

Replace https://github.com/pluginsGLPI/fields/pull/670

### DIFF
--- a/hook.php
+++ b/hook.php
@@ -28,6 +28,8 @@
  * -------------------------------------------------------------------------
  */
 
+use Glpi\Api\HL\Doc\Schema;
+
 /**
  * Plugin install process
  *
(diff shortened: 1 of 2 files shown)


stonebuzz:
@cconard96 

I'd like your opinion on this.

Thank you in advance

cconard96:
I don't have the plugin installed/configured anymore, but I didn't see anything in the code itself of concern as long as there is no way for the fields plugin to add fields to native custom assets and custom dropdowns. Both add custom fields to a "custom_fields" property and it doesn't seem like this code checks if it already exists to merge the properties instead.

smurdza:
@stonebuzz do you know when this implement will be ready ?

PR #1175, opened 2026-04-01
fix: Fixed field deletion blocked when at least one form question use another field

ccailly:

- [X] I have performed a self-review of my code.
- [X] I have added tests (when available) that prove my fix is effective or that my feature works.
- [X] I have updated the CHANGELOG with a short functional description of the fix or new feature.

## Description

- #1173 must be merged before
- It fixes !43103

Currently, if a question is of the `Field` type and uses a field, it is no longer possible to delete any field, regardless of its parent block.
This PR fixes this behavior by ensuring that deletion is blocked only for fields that are actually used in a form.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fix error when submitting a form with an hidden question of type `Field`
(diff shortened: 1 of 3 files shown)


stonebuzz (approved):

Rom1-B (approved):

stonebuzz:
@ccailly 
can you rebase ?

PR #1179, opened 2026-04-13
Fix - Avoid SQL warning when no user session is active during plugin init

RomainLvr:

## Description

- It fixes !43370

When initializing the plugin during boot (before any user session), `getEntitiesRestrictCriteria()` was called without an active session, producing `entities_id = ''` in the SQL query and triggering a mass MySQL warning (1292: Truncated incorrect DECIMAL value).

Fix by skipping entity restriction when no user is logged in (`!Session::getLoginUserID()`), consistent with the existing check already in place in `setup.php`.

Error : 
```php
SQL: SELECT * FROM `glpi_plugin_fields_containers` WHERE `is_active` = '1' AND `type` = 'dom' AND ((`itemtypes` LIKE '%\\\"Ticket\\\"%') OR (`itemtypes` LIKE '%\\\"Change\\\"%') OR (`itemtypes` LIKE '%\\\"Problem\\\"%')) AND (`glpi_plugin_fields_containers`.`entities_id` = '') ORDER BY `name`
  Warnings: 
1292: Truncated incorrect DECIMAL value: '' at DBmysql.php line 444
  Backtrace :
  ./src/DBmysql.php:444                              
  ./src/DBmysqlIterator.php:129                      DBmysql->doQuery()
  ./src/DBmysql.php:1088                             DBmysqlIterator->execute()
  ./src/CommonDBTM.php:632                           DBmysql->request()
  ./plugins/fields/inc/questiontype.class.php:384    CommonDBTM->find()
  ./plugins/fields/inc/questiontype.class.php:426    PluginFieldsQuestionType->getAvailableBlocks()
  ./plugins/fields/setup.php:428                     PluginFieldsQuestionType::hasAvailableFields()
  ./plugins/fields/setup.php:136                     plugin_fields_register_plugin_types()
  ./src/Plugin.php:475                               plugin_init_fields()
  ./src/Plugin.php:428                               Plugin::load()
  ...tener/PostBootListener/InitializePlugins.php:75 Plugin->init()
  .../event-dispatcher/Debug/WrappedListener.php:116 Glpi\Kernel\Listener\PostBootListener\InitializePlugins->onPostBoot()
  ...ymfony/event-dispatcher/EventDispatcher.php:220 Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke()
  ...symfony/event-dispatcher/EventDispatcher.php:56 Symfony\Component\EventDispatcher\EventDispatcher->callListeners()
  ...spatcher/Debug/TraceableEventDispatcher.php:142 Symfony\Component\EventDispatcher\EventDispatcher->dispatch()
  ./src/Glpi/Kernel/Kernel.php:149                   Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher->dispatch()
  ./vendor/symfony/http-kernel/Kernel.php:201        Glpi\Kernel\Kernel->boot()
  ./public/index.php:71                              Symfony\Component\HttpKernel\Kernel->handle()


```
Close https://github.com/pluginsGLPI/fields/issues/1172

### DIFF
--- a/inc/questiontype.class.php
+++ b/inc/questiontype.class.php
@@ -380,7 +380,17 @@ final class PluginFieldsQuestionType extends AbstractQuestionType implements For
         $field_container  = new PluginFieldsContainer();
         $available_blocks = [];
 
-        $entity_restrict = (isCommandLine() || !Session::getLoginUserID()) ? [] : getEntitiesRestrictCriteria(PluginFieldsContainer::getTable(), '', '', true);
+        $entity_restrict = isCommandLine() ? [] : getEntitiesRestrictCriteria(PluginFieldsContainer::getTable(), '', '', true);
+
+        // If entity restriction contains an empty string value, it means no valid session
+        // is active. Running the query would produce a MySQL
+        // warning (1292: Truncated incorrect DECIMAL value)
+        foreach ($entity_restrict as $criterion) {
+            if (is_array($criterion) && in_array('', $criterion, true)) {
+                return $available_blocks;
+            }
+        }
+
         $result           = $field_container->find([
             'is_active' => 1,
             'type'      => 'dom',
--- a/inc/questiontype.class.php
+++ b/inc/questiontype.class.php
@@ -380,7 +380,7 @@ final class PluginFieldsQuestionType extends AbstractQuestionType implements For
         $field_container  = new PluginFieldsContainer();
         $available_blocks = [];
 
-        $entity_restrict = isCommandLine() ? [] : getEntitiesRestrictCriteria(PluginFieldsContainer::getTable(), '', '', true);
+        $entity_restrict = (isCommandLine() || !Session::getLoginUserID()) ? [] : getEntitiesRestrictCriteria(PluginFieldsContainer::getTable(), '', '', true);
         $result           = $field_container->find([
             'is_active' => 1,
             'type'      => 'dom',
--- a/inc/questiontype.class.php
+++ b/inc/questiontype.class.php
@@ -380,16 +380,14 @@ final class PluginFieldsQuestionType extends AbstractQuestionType implements For
         $field_container  = new PluginFieldsContainer();
         $available_blocks = [];
 
-        $entity_restrict = isCommandLine() ? [] : getEntitiesRestrictCriteria(PluginFieldsContainer::getTable(), '', '', true);
-
-        // If entity restriction contains an empty string value, it means no valid session
-        // is active. Running the query would produce a MySQL
-        // warning (1292: Truncated incorrect DECIMAL value)
-        foreach ($entity_restrict as $criterion) {
-            if (is_array($criterion) && in_array('', $criterion, true)) {
-                return $available_blocks;
-            }
+        // No session and not CLI: return early to avoid invalid SQL criterion from
+        // getEntitiesRestrictCriteria() (returns entities_id = '' on integer column,
+        // producing MySQL warning 1292). Consistent with the guard in setup.php.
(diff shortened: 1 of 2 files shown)


inc/questiontype.class.php
```diff
@@ -380,7 +380,7 @@ private function getAvailableBlocks(): array
         $field_container  = new PluginFieldsContainer();
         $available_blocks = [];
 
-        $entity_restrict = isCommandLine() ? [] : getEntitiesRestrictCriteria(PluginFieldsContainer::getTable(), '', '', true);
+        $entity_restrict = (isCommandLine() || !Session::getLoginUserID()) ? [] : getEntitiesRestrictCriteria(PluginFieldsContainer::getTable(), '', '', true);
```
Rom1-B: Normally, you're either in CLI mode or logged in as an authenticated user. It seems to be neither, so I don't think this should be allowed, to avoid creating a security issue

Rom1-B (changes requested):

inc/questiontype.class.php
```diff
@@ -381,6 +381,16 @@ private function getAvailableBlocks(): array
...
+
+        // If entity restriction contains an empty string value, it means no valid session
+        // is active. Running the query would produce a MySQL
+        // warning (1292: Truncated incorrect DECIMAL value)
+        foreach ($entity_restrict as $criterion) {
+            if (is_array($criterion) && in_array('', $criterion, true)) {
+                return $available_blocks;
+            }
+        }
+
```
Rom1-B: Maybe something like this:
```suggestion
        $activeentities = getEntitiesRestrictCriteria(PluginFieldsContainer::getTable(), '', '', true);
        if ($activeentities === []) {
            return [];
        }
        $entity_restrict = isCommandLine() ? [] : $activeentities;
```

RomainLvr: `getEntitiesRestrictCriteria` returns an array of criteria. The problem here isn't that `getEntitiesRestrictCriteria` returns an empty array; rather, it returns the criterion `glpi_plugin_fields_containers.entities_id = ""`, so the suggestion won't fix the problem.

Rom1-B: That's true, but ultimately I think the problem lies in the `getEntitiesRestrictCriteria()` function, which returns an invalid value. Instead of returning `...entities_id = ''` (which can never work), it should return `[new QueryExpression(‘false’)];`
_For your information, it also checks `isCommandLine()`, so the one here will likely be redundant._

RomainLvr: You're right, the root cause is actually in the GLPI core — `DbUtils::getEntitiesRestrictCriteria()` in `DbUtils.php`.

When called with no active session, no CLI context, and no cron context, the function falls through all conditions without an `else` branch, leaving `$value` as an empty string `''`. This produces an invalid SQL criterion `entities_id = ''` on an integer column, which triggers MySQL warning.

I've updated this with a cleaner guard:

```php
if (!Session::getLoginUserID() && !isCommandLine()) {
    return $available_blocks;
}
```

This is semantically explicit (no session = no available blocks), consistent with the existing guard already in place in `setup.php` l.138 — it doesn't rely on inspecting the internal structure of the criteria returned by core.

The real fix should live in the core: adding an `else { return [new QueryExpression('false')]; }` branch to `getEntitiesRestrictCriteria()` so all callers are protected. I'll open a separate PR on the GLPI core for that.

Rom1-B (approved):

inc/questiontype.class.php
RomainLvr: here is the fix for the core : https://github.com/glpi-project/glpi/pull/23934

PR #1185, opened 2026-04-22
Fix: table rename aborts when GenericObject migration exceeds MySQL's…

Rom1-B:
… 64-char limit

## Description

- It fixes !43610

When migrating containers from GenericObject to GlpiCustomAsset, the plugin renames the underlying DB table. For certain object names, the new name exceeds MySQL's 64-character identifier limit, aborting the     
  migration with error 1059.                                                                                                                                                                                          
                                                                                                                                                                                                                      
  A naive table name truncation would not suffice: GLPI derives table names from class names at runtime, so the stored container name and the DB table must stay in sync. The fix truncates the container name until  
  the generated table fits within 64 chars, then persists the shortened name to the DB — mirroring an identical safeguard already present in installUserData.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [UNRELEASED]
+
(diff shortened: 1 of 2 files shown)


Rom1-B:
@stonebuzz 
The customer has approved the solution

stonebuzz (approved):

PR #1187, opened 2026-04-24
Fix(Core): Optimize container loading when there are a large number of entities

stonebuzz:
…f entities

## Description

- It fixes #42749

A user with nearly 5,000 entities is experiencing slowness when opening a ticket containing a container.

They are using a tree view with recursive mode enabled.

Currently, the plugin adds a WHERE clause on entities using all the user’s entities (`glpi_active_entities`), which significantly impacts performance.

The logic used to verify whether access to the container is authorized has been moved further downstream in order to lighten the SQL query.

It now relies on `Session::haveAccessToEntity`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fix text area fields size and alignment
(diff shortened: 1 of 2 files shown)


Rom1-B (approved):

stonebuzz:
validated by customer

PR #1195, opened 2026-05-11
Adding a verification of obj value if not an array so we don't try to use push on a empty string

emitlinks:

- [ x ] I have performed a self-review of my code.

## Description

- It fixes #1155 
- Here is a brief description of what this PR does : 
It checks the value of the item that has multiple matches in obj to see if it's an empty string instead of an empty array (this can happen when a multi-value field is checked) in which case it's replaced by an empty array so the code doesn't bug when trying to push a value to a string.


stonebuzz (commented):
Can you adapt `CHANGELOG.md`

stonebuzz (approved):

Rom1-B (approved):

PR #1197, opened 2026-05-15
fix: nested array stored in DB for readonly dropdown

TarekRemo:

## Description

- It fixes ticket !44062.

- Steps to reproduce:
  - Create a container with at least two multiple-dropdown custom fields : one set as readonly and one editable.
  - Open an item that has a value saved for the readonly dropdown (e.g. ["3"]).
  - Change the value of the editable dropdown and save the form.
  - Reload the page : a PHP error is thrown.
  
- Root cause:
  - The `dropdownField` Twig macro pre-appends `[]` to the field name for all multiple inputs (e.g. `plugin_fields_donefielddropdowns_id[]`). 
  - For readonly fields, `Dropdown::show()` renders hidden inputs and appends another `[]`, producing `plugin_fields_donefielddropdowns_id[][]`. 
  - When the form is saved, PHP parses this double-bracketed name as a nested array (`[['3']]` instead of `['3']`), which gets JSON-encoded as `[["3"]]` and stored corrupted in the database. 
  - On the next page load, `Dropdown::show()` receives the decoded nested array, iterates over it, and tries to use an array as an array key inside isset(), triggering a PHP 8 TypeError.
  
- Fix:
  - Before json_encode-ing a multiple field value for storage, `flattenScalars()` collapses any nested arrays into a flat list of scalars, preventing the corruption from ever reaching the database.
  - The same flattening is applied at display time so that values already corrupted in the database before this fix are handled.

### DIFF
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -1413,11 +1413,12 @@ HTML;
                     // Add new values to existing ones
                     $existing_values = json_decode($obj->fields[$field_name] ?? '[]', true);
                     $new_values      = is_array($data[$field_name]) ? $data[$field_name] : [$data[$field_name]];
+                    $new_values      = $this->flattenScalars($new_values);
                     $data[$field_name] = json_encode(array_values(array_unique(array_merge($existing_values, $new_values))));
 
                 } else {
                     $value = $data[$field_name];
-                    $value = is_array($value) ? $value : [];
+                    $value = is_array($value) ? $this->flattenScalars($value) : [];
                     $data[$field_name] = json_encode($value);
                 }
             } elseif (array_key_exists('_' . $field_name . '_defined', $data)) {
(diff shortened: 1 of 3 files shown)


inc/container.class.php
```diff
@@ -1966,6 +1967,29 @@ public static function preItem(CommonDBTM $item)
...
+                        $flat[] = $inner;
+                    }
+                }
+            } elseif (is_scalar($v)) {
+                $flat[] = $v;
+            }
+        }
+
+        return $flat;
+    }
```
Rom1-B, line 1984: Maybe easier to read:
```suggestion
    private function flattenScalars(array $values): array
    {
        $flat = [];
        foreach ($values as $v) {
            array_push($flat, ...(is_array($v) ? $v : [$v]));
        }
        return array_filter($flat, is_scalar(...));
    }
```

TarekRemo: done

Rom1-B (approved):

stonebuzz (approved):

Jerome-Gdrs:
Hi. Do you know when this will be released ? Thanks :)

PR #1212, opened 2026-06-22
fix(container.class.php): fix the name of dropdown fields in log

TarekRemo:

## Description

- It fixes !44835
- Before this fix, the name of dropdown fields was missing in logs.
- For dropdown fields (both custom and `dropdown-<ItemType>`), the search option sets: 
  - `field` → `completename` or `name` (from the foreign table, e.g. `glpi_users.name`)
  - `linkfield` → the actual column in the container table (e.g. `usersfield_0`)
- So, the current code compares `name` against the actual DB column key when updating = it never matches -> the name of the field doesn't appear.
- fix: add `$searchoption['linkfield'] == $key` to check the the actual column in the container table.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fix the name of dropdown fields in logs when updating an item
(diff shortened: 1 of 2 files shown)


stonebuzz (approved):

stonebuzz:
Awaiting customers feedback

PR #1214, opened 2026-06-25
fix: apply translation to container tab name in getTabNameForItem

bsitfeng:
Container tab names were displayed using \$data['label'] directly without going through the translation system. This fix applies getLabelFor() to correctly display translated tab labels based on the user's language.

Fixes tab name translation not working when language is set to non-default.

### DIFF
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -1337,7 +1337,7 @@ HTML;
                         if (!$item->isEntityAssign() || in_array($item->fields['entities_id'], $entities)) {
                             $display_condition = new PluginFieldsContainerDisplayCondition();
                             if ($display_condition->computeDisplayContainer($item, $data['id'])) {
-                                $tabs_entries[$data['id']] = self::createTabEntry($data['label'], 0, null, PluginFieldsContainer::getIcon());
+                                $tabs_entries[$data['id']] = self::createTabEntry(PluginFieldsLabelTranslation::getLabelFor(array_merge($data, ['itemtype' => 'PluginFieldsContainer'])), 0, null, PluginFieldsContainer::getIcon());
                             }
                         }
                     }


stonebuzz:
Hi @bsitfeng 

can you adapt CHANGELOG.md

```markdown
## [UNRELEASED]

### Fixed

- Container tab names are now correctly translated
```

inc/container.class.php
```diff
@@ -1337,7 +1337,7 @@ public function getTabNameForItem(CommonGLPI $item, $withtemplate = 0)
                         if (!$item->isEntityAssign() || in_array($item->fields['entities_id'], $entities)) {
                             $display_condition = new PluginFieldsContainerDisplayCondition();
                             if ($display_condition->computeDisplayContainer($item, $data['id'])) {
-                                $tabs_entries[$data['id']] = self::createTabEntry($data['label'], 0, null, PluginFieldsContainer::getIcon());
+                                $tabs_entries[$data['id']] = self::createTabEntry(PluginFieldsLabelTranslation::getLabelFor(array_merge($data, ['itemtype' => 'PluginFieldsContainer'])), 0, null, PluginFieldsContainer::getIcon());
```
stonebuzz, line 1340:
```suggestion
                                $label = PluginFieldsLabelTranslation::getLabelFor(
                                    array_merge($data, ['itemtype' => PluginFieldsContainer::class])
                                );
                                $tabs_entries[$data['id']] = self::createTabEntry($label, 0, null, PluginFieldsContainer::getIcon());
```

improves readability

stonebuzz (changes requested):

inc/container.class.php
stonebuzz: `getTabNameForItem` has no test in `tests/Units/`. Can you add a test case where a container has a translation for the active session language and verify the translated label is returned, and another where no translation exists and the raw `label` value is returned as fallback?

Rom1-B: Redundant lookup: `$tab_label` (loop var, line 1327) already holds this same translated value from `getEntries('tab', true)`. Reuse it instead of recomputing (extra query + hardcoded `'PluginFieldsContainer'` string).

```suggestion
                                $tabs_entries[$data['id']] = self::createTabEntry($tab_label, 0, null, PluginFieldsContainer::getIcon());
```
---
No test for `getTabNameForItem()`. Cover: translated label returned when a translation exists, raw `label` returned as fallback otherwise.

Rom1-B (changes requested):

PR #1221, opened 2026-07-06
Fix(Migration): prevent container name corruption on partial update

stonebuzz:

## Description

During plugin installation when migrating containers from GenericObject to
CustomAsset, containers end up with a corrupted `name` field (e.g.
`sixonefourfourthreesixtwozeroeight`) and an empty `label`, causing a
subsequent MySQL error when the dynamic table name exceeds 64 characters.

### Root cause

`installBaseData` calls `update()` with only `id` and `itemtypes`  (no `label` key  to update) see L249

`prepareInputForUpdate()` called `prepareLabel()` unconditionally, which:

1. passed `''` to `getSystemNameFromLabel()`, which fell back to
   `random_int(0, mt_getrandmax())` for an empty name
2. passed that integer through `replaceIntByLetters()`, spelling each
   digit out (`6 1 4 4 3 6 2 0 0 8`  =>  `sixonefourfourthreesixtwozeroeight`)

### State before migration

mysql> SELECT id, name, label, itemtypes FROM glpi_plugin_fields_c
+----+---------+---------+--------------------------------+
| id | name    | label   | itemtypes                      |
+----+---------+---------+--------------------------------+
|  3 | contain | contain | ["PluginGenericobjectContain"] |
+----+---------+---------+--------------------------------+

### State after migration (buggy)

+----+------------------------------------+-------+-------------------------------------+
| id | name                               | label | itemtypes
+----+------------------------------------+-------+-------------------------------------+
|  3 | sixonefourfourthreesixtwozeroeight |       | ["Glpi\CustomA
+----+------------------------------------+-------+-------------------------------------+

This triggers:

RuntimeException: MySQL query error: Identifier name
'glpi_plugin_fields_glpicustomassetcontainassetsixonefourfourthree
is too long (1059)

### DIFF
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -648,8 +648,10 @@ class PluginFieldsContainer extends CommonDBTM
 
     public function prepareInputForUpdate($input)
     {
-        unset($input['label']);
-        unset($input['name']);
+        if (isAPI() || isset($input['form_submission']) ) {
+            unset($input['label']);
+            unset($input['name']);
+        }
 
         return $input;
     }
@@ -973,17 +975,24 @@ HTML;
         echo '<tr>';
         echo "<td width='20%'>" . __('Label') . ' : </td>';
         echo "<td width='30%'>";
-        $options = [
+        $label_options = [
             'value' => $this->fields['label'],
         ];
 
         if (!$this->isNewID($ID)) {
-            $options['readonly'] = true;
+            $label_options['readonly'] = true;
         }
 
         echo Html::input(
             'label',
-            $options,
+            $label_options,
+        );
+
+        echo Html::hidden(
+            'form_submission',
+            [
+                'value' => true,
+            ],
         );
         echo '</td>';
         echo "<td width='20%'>&nbsp;</td>";
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -648,7 +648,10 @@ class PluginFieldsContainer extends CommonDBTM
 
     public function prepareInputForUpdate($input)
     {
-        return PluginFieldsToolbox::prepareLabel($input);
+        unset($input['label']);
+        unset($input['name']);
+
+        return $input;
     }
 
     public function prepareInputForAdd($input)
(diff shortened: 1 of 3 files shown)


stonebuzz:
Currently, updating the **label** and/or **name** triggers the regeneration of the associated database tables and generated files. This behavior is not desirable and was not possible in previous versions. I'm not sure when this change was introduced, but it should be reverted.

inc/container.class.php
```diff
@@ -648,7 +648,10 @@ public function defineTabs($options = [])
 
     public function prepareInputForUpdate($input)
     {
-        return PluginFieldsToolbox::prepareLabel($input);
+        unset($input['label']);
+        unset($input['name']);
+
+        return $input;
     }
```
Rom1-B, line 657: Blanket strip of `name`/`label` also hits the plugin's own migrations, which rely on `update()` to persist a corrected `name`:
- `installBaseData()` L249: renames the table, but the DB `name` write is now dropped → row points at a table that no longer exists.
- `installUserData()` L341: the "fix pre-1.9.2 numeric names" migration — its whole point was persisting the corrected `name`. Now silently no-ops, reintroducing the corruption this PR fixes.

Scope the strip to form submissions only (dedicated internal update path, or a flag), not all `update()` calls.

---

No test covers `prepareInputForUpdate()` stripping `name`/`label`, or the migration paths above.

inc/container.class.php
```diff
@@ -970,11 +973,17 @@ public function showForm($ID, $options = [])
...
+        if (!$this->isNewID($ID)) {
+            $options['readonly'] = true;
+        }
+
         echo Html::input(
             'label',
-            [
-                'value' => $this->fields['label'],
-            ],
+            $options,
```
Rom1-B: Reused `$options` (the `showForm($ID, $options)` param) for the label input, clobbering it before `showFormButtons($options)` at L1089. Harmless today only because the current caller passes no extra keys. Rename to `$label_options`.


```suggestion
        $label_options = [
            'value' => $this->fields['label'],
        ];

        if (!$this->isNewID($ID)) {
            $label_options['readonly'] = true;
        }

echo Html::input(
'label',
            $label_options,
```

Rom1-B (changes requested):

stonebuzz:
close in favor of 

https://github.com/pluginsGLPI/fields/pull/1225

PR #1222, opened 2026-07-08
Fix: search crash on duplicate field names

Rom1-B:

## Description

- It fixes !45192
- Two containers can legitimately define a dropdown field with the same name, so both share the same underlying dropdown table.
- Searching on the related itemtype then crashed with a Twig runtime error, because the plugin looked up the field by name instead of by its unique id, and the lookup found two matching rows instead of one.

```
An exception thrown during rendering template ("`PluginFieldsField::getFromDBByCrit()` expects get one result, 2 found in query "SELECT `id` FROM `glpi_plugin_fields_fields` WHERE `name` = 'accessoirefield' AND `multiple` = '1'".") in "pages/generic_list.html.twig" line 37.
```

- The field is now identified by its id, which is already computed when the search options are built and used elsewhere in the plugin for the same purpose.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [unreleased]
 
+### Fixed
+
+- Fix search crash when two containers share a dropdown field with the same name.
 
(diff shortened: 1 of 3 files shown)


stonebuzz (approved):
LGTM

PR #1229, opened 2026-07-21
fix(container.class): Prevent item creation with null value for mandatory fields

TarekRemo:

## Description

- It fixes !45283
- Before this fix, in `preItem()`, the validation of required fields (`validateValues()`) was only called if `populateData()` did not return `false`. However, `populateData()` returns `false` when none of the block's fields are present in the input. Result: A creation that did not submit the fields (API/CLI, or a form without mapping) completely bypassed the check and created the element with a `NULL` value in a required field.
- Fix:  On creation only (`isNewItem()`), when `populateData()` returns `false` and the continer type is `dom`, `validateValues()` is still called on an empty dataset, which causes the missing required fields to fail validation and blocks creation.

### DIFF
--- a/inc/container.class.php
+++ b/inc/container.class.php
@@ -1970,12 +1970,19 @@ HTML;
             return true;
         }
 
+        //call validateValues() with a minimal data array to check for missing mandatory fields
+        //in case populateData() fails
         if ($item->isNewItem() && $loc_c->fields['type'] === 'dom') {
+            $status_field_name = PluginFieldsStatusOverride::getStatusFieldName($item::getType());
             $data = ['plugin_fields_containers_id' => $c_id];
+            if (array_key_exists($status_field_name, $item->input) && $item->input[$status_field_name] !== '') {
+                $data[$status_field_name] = (int) $item->input[$status_field_name];
+            } elseif (array_key_exists($status_field_name, $item->fields) && $item->fields[$status_field_name] !== '') {
+                $data[$status_field_name] = (int) $item->fields[$status_field_name];
+            }
+
             if (self::validateValues($data, $item::getType(), isset($_REQUEST['massiveaction'])) === false) {
                 $item->input = [];
-
-                return false;
             }
         }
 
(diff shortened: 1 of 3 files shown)


inc/container.class.php
```diff
@@ -1970,6 +1970,15 @@ public static function preItem(CommonDBTM $item)
...
         }
 
+        if ($item->isNewItem() && $loc_c->fields['type'] === 'dom') {
+            $data = ['plugin_fields_containers_id' => $c_id];
+            if (self::validateValues($data, $item::getType(), isset($_REQUEST['massiveaction'])) === false) {
+                $item->input = [];
+
+                return false;
+            }
+        }
```
Rom1-B, line 1987: The minimal `$data` array built for this new validation path omits the status field. `populateData()`'s normal path always sets `$data[$status_field_name]` from `$item->input`/`$item->fields` (lines 2029-2034), and `validateValues()` resolves `$status_value = $data[$status_field_name] ?? null` to apply `PluginFieldsStatusOverride` mandatory/readonly relaxations (line 1673). Since this new branch never sets that key, any status override waiving a field's mandatory flag for a given creation status is silently ignored here, so a legitimate creation (API/CLI, `dom` container, status override in place) gets incorrectly blocked.

```suggestion
        if ($item->isNewItem() && $loc_c->fields['type'] === 'dom') {
            $status_field_name = PluginFieldsStatusOverride::getStatusFieldName($item::getType());
            $data = ['plugin_fields_containers_id' => $c_id];
            if (array_key_exists($status_field_name, $item->input) && $item->input[$status_field_name] !== '') {
                $data[$status_field_name] = (int) $item->input[$status_field_name];
            } elseif (array_key_exists($status_field_name, $item->fields) && $item->fields[$status_field_name] !== '') {
                $data[$status_field_name] = (int) $item->fields[$status_field_name];
            }
            if (self::validateValues($data, $item::getType(), isset($_REQUEST['massiveaction'])) === false) {
                $item->input = [];

                return false;
            }
        }
```

Rom1-B (changes requested):

inc/container.class.php
TarekRemo: done

Rom1-B (approved):

PR #1230, opened 2026-07-27
fix handle native glpi dropdown types when binding additional field to form destination

Herafia:

## Description

- It fixes !45483
- Fixes the submission of a form containing an additional field of the native glpi dropdown type (e.g : Groups) linked to a destination. The raw response (itemtype/items_id array) was inserted into the database, causing an SQL error and an error when rendering the response.
We now extract the items_id for all dropdown types, and the id is cast to int before being read.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fix item creation with null value for mandatory fields
 - Fix search crash when two containers share a dropdown field with the same name.
+- fix handle native GLPI dropdown types when binding additional fields to form destination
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fix item creation with null value for mandatory fields
 - Fix search crash when two containers share a dropdown field with the same name.
+- fix handle native GLPI dropdown types when binding additional fields to form destination
```
Rom1-B:
```suggestion
- Fix handle native GLPI dropdown types when binding additional fields to form destination
```

Rom1-B (approved):

PR #1235, opened 2026-08-01
feat(forms): implement embeddable rendering for Fields question type

UncleTomsCabi:
## Summary

Implement the new embeddable question type interface for the Fields question type.

This allows the Fields question type to be rendered inside composite question types while keeping all rendering logic inside the Fields plugin.

## What is implemented

- implementation of the new embeddable interface;
- embedded administration rendering;
- embedded end-user rendering;
- automatic block detection;
- selection of the concrete field only;
- reuse of the existing Fields rendering logic;
- full compatibility with existing standalone Fields questions.

## Motivation

The Fields plugin already contains all rendering logic required to display its controls.

Instead of duplicating this logic in other plugins, it is now exposed through the common interface introduced in GLPI.

This keeps responsibilities separated:

- Fields knows how to render Fields controls;
- composite question types only request rendering.

## Backward compatibility

No existing behaviour is changed.

Standalone Fields questions continue to work exactly as before.

## Related

Depends on:

- GLPI Core PR: https://github.com/glpi-project/glpi/pull/25065

Used by:

- Advanced Forms PR: https://github.com/pluginsGLPI/advancedforms/pull/47

### DIFF
--- a/inc/questiontype.class.php
+++ b/inc/questiontype.class.php
@@ -37,11 +37,14 @@ use Glpi\Form\Migration\FormQuestionDataConverterInterface;
 use Glpi\Form\Question;
 use Glpi\Form\QuestionType\AbstractQuestionType;
 use Glpi\Form\QuestionType\QuestionTypeCategoryInterface;
+use Glpi\Form\QuestionType\QuestionTypeEmbeddableInterface;
 
 use function Safe\json_decode;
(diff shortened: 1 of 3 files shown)


Rom1-B:
I'm saving the PR as a draft because it depends on others that haven't been approved yet.

PR #1239, opened 2026-08-06
Fix: trigger GenericObject dropdown fields migration

Rom1-B:

## Description

- It fixes !43878
- The migration guard that detects legacy GenericObject dropdown fields read the wrong result key from the count query, so it always evaluated to false.
- As a result, GenericObject dropdown fields were never migrated to CustomAsset dropdown fields, and affected fields displayed the wrong label on item forms.


MyvTsv (approved):

MyvTsv:
That looks good to me. I'll wait for the client's feedback before approving the fix.

PR #974, opened 2025-07-07
fix(MassiveAction): fix incompatibility with multiple dropdowns

MyvTsv:
## Checklist before requesting a review

*Please delete options that are not relevant.*

- [x] I have performed a self-review of my code.
- [x] I have added tests (when available) that prove my fix is effective or that my feature works.
- [x] I have updated the CHANGELOG with a short functional description of the fix or new feature.
- [ ] This change requires a documentation update.

## Description

- It fixes !38252
- Massive actions did not work on the multiple fields of the field plugin

## Screenshots (if appropriate):
Notification

PR #1247, opened 2026-08-20
Fix - Massive action update on CustomAssets

RomainLvr:

## Description

- It fixes !45865

Massive action "Update" was broken for Fields-plugin fields attached to Custom Assets:

1. **The value input never displayed.** `PluginFieldsField::showSingle()` looks up the field's container with a `LIKE` query against `containers.itemtypes`, but didn't escape backslashes the same way `PluginFieldsContainer::getAddSearchOptions()` already does. Since CustomAsset itemtypes are namespaced classes (e.g. `Glpi\CustomAsset\Transmission`), the query never matched, so the field never rendered, only the "Send" button showed, and submitting produced an `Undefined array key` warning.
2. **`hook.php` masked the failure.** `plugin_fields_MassiveActionsFieldsDisplay()` ignored `showSingle()`'s return value and always returned `true`, preventing GLPI core's fallback renderer from ever kicking in.
3. **Multi-select "GLPI item" dropdown fields got wiped.** A field of type `dropdown-<Itemtype>` (e.g. a multi-select of Locations) with `multiple` enabled was silently emptied whenever a *different* field in the same container was mass-updated afterwards. The plain `dropdown` type already had a guard against this (fixed in #795 / #974), but it was missing on this sibling code path.

### Fix

- Apply the same backslash-escaping logic in `showSingle()` as in `getAddSearchOptions()`.
- Make `hook.php` return `showSingle()`'s actual result.
- Apply the same "don't blank an untouched field" guard to the `dropdown-<Itemtype>` multi-value branch in `PluginFieldsContainer::populateData()`.

### Test plan

- **The test only passes if the patch https://github.com/glpi-project/glpi/pull/25210 is applied otherwise, the error ‘Converting an array to a string’ causes the tests to fail.**

- Added regression tests :
  - `MassiveActionCustomAssetTest` => field display for CustomAssets.
  - `MassiveActionMultipleDropdownTest` / `MassiveActionRealFlowTest` => plain `dropdown` multiple field survives a sibling field's massive update.
  - `MassiveActionGlpiItemDropdownTest` => `dropdown-<Itemtype>` multiple field survives a sibling field's massive update (fails without the fix, passes with it).

_PS : i've added psalm.xml because it was missing from `make test` process_


Rom1-B (changes requested):
Please fix CI (phpunit)

inc/container.class.php
```diff
@@ -2097,12 +2097,22 @@ private static function populateData($c_id, CommonDBTM $item)
...
+                        isset($item->input[$defined_key])
+                        && $item->input[$defined_key]
+                    ) { //multi dropdown is empty or has been emptied
                         $data[$field['name']] = [];
+                        $has_fields           = true;
+                    } elseif (isset($_REQUEST['massiveaction'])) { // called from massiveaction
+                        if (isset($_POST[$field['name']])) {
+                            $data[$field['name']] = $_POST[$field['name']];
+                            $has_fields           = true;
+                        }
```
Rom1-B:
```suggestion
                    } elseif (
                        isset($_REQUEST['massiveaction'])
                        && (isset($_POST[$field['name']])
                    ) {
                        $data[$field['name']] = $_POST[$field['name']];
                        $has_fields           = true;
```

psalm.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<psalm
+    runTaintAnalysis="true"
+>
+    <projectFiles>
+        <directory name="src"/>
+        <file name="hook.php"/>
+        <file name="setup.php"/>
```
Rom1-B: I think we need to create a dedicated PR for PSALM to make it complete:
```suggestion
        <directory name="ajax" />
        <directory name="front" />
        <directory name="inc" />
        <directory name="src" />
        <directory name="tests" />
        <file name="hook.php" />
        <file name="setup.php" />
```

inc/field.class.php
```diff
@@ -1361,6 +1361,14 @@ public static function showSingle($itemtype, $searchOption, $massiveaction = fal
             (string) $searchOption['linkfield'],
         );
 
+        // itemtype is stored in a JSON array, so entry is surrounded by double quotes
+        $search_string = json_encode($itemtype);
+        // Backslashes must be doubled in LIKE clause according to MySQL documentation
+        // But do not escape backslashes for CustomAsset, as they are alrady escaped
+        if (!str_contains($search_string, 'CustomAsset')) {
```
Rom1-B:
```suggestion
        // But do not escape backslashes for namespaced itemtypes, as they are already escaped
        if (!str_contains((string) $itemtype, '\\')) {
```

Rom1-B (changes requested):

stonebuzz (changes requested):
see review

Rom1-B (approved):

stonebuzz (approved):

stonebuzz:
Pending the requester’s feedback.

Could you ask them to test the patch and confirm whether the issue is resolved?

PR #1248, opened 2026-08-21
Feat: rename oversized containers from their edit form

Rom1-B:

- [x] This change requires a documentation update.

## Description

- It fixes !45901
- When a container's computed table name exceeds MySQL's 64-character limit, the plugin now disables it instead of crashing, and shows a rename form directly on the container's edit page.
- Submitting a shorter name renames the container and reactivates it; if a table with matching data already exists under that name (e.g. after a version migration), its data is preserved instead of being replaced by an empty table.
- `plugins:fields:check_database` now also reports container/item type pairs with no matching table, and tables with no matching container/item type pair.

### DIFF
--- a/inc/checkdatabasecommand.class.php
+++ b/inc/checkdatabasecommand.class.php
@@ -114,7 +114,7 @@ class PluginFieldsCheckDatabaseCommand extends AbstractCommand
             );
             foreach ($tables_consistency['missing'] as $entry) {
                 $output->writeln(sprintf(
-                    '-> container #%d (%s): expected table "%s" not found',
+                    __('-> container #%d (%s): expected table "%s" not found', 'fields'),
                     $entry['container_id'],
                     $entry['itemtype'],
                     $entry['table'],
--- a/inc/checkdatabasecommand.class.php
+++ b/inc/checkdatabasecommand.class.php
@@ -47,7 +47,9 @@ class PluginFieldsCheckDatabaseCommand extends AbstractCommand
                 __('- some deleted fields may still be present in database (bug introduced in version %s and fixed in version %s)', 'fields'),
                 '1.15.0',
                 '1.15.3',
-            ),
+            )
+            . "\n"
+            . __('- container/item type pairs with no matching table, or tables with no matching container/item type pair', 'fields'),
         );
 
         $this->addOption(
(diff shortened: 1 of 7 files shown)


inc/checkdatabasecommand.class.php
```diff
@@ -66,41 +68,68 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                '<error>' . sprintf(__('%d container/item type pair(s) have no matching table.', 'fields'), count($tables_consistency['missing'])) . '</error>',
+                OutputInterface::VERBOSITY_QUIET,
             );
+            foreach ($tables_consistency['missing'] as $entry) {
+                $output->writeln(sprintf(
+                    '-> container #%d (%s): expected table "%s" not found',
+                    $entry['container_id'],
+                    $entry['itemtype'],
+                    $entry['table'],
+                ));
```
stonebuzz, line 121: inconsistent with the dead-fields block above.

```suggestion
                $output->writeln(sprintf(
                    __('-> container #%d (%s): expected table "%s" not found', 'fields'),
                    $entry['container_id'],
                    $entry['itemtype'],
                    $entry['table'],
                ));
```

inc/checkdatabasecommand.class.php
```diff
@@ -66,41 +68,68 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        }
+
+        if ($tables_consistency['orphaned'] !== []) {
             $output->writeln(
-                '<comment>' . $next_command . '</comment>',
+                '<error>' . sprintf(__('%d table(s) do not match any container/item type pair.', 'fields'), count($tables_consistency['orphaned'])) . '</error>',
                 OutputInterface::VERBOSITY_QUIET,
             );
+            foreach ($tables_consistency['orphaned'] as $table) {
+                $output->writeln(sprintf('-> "%s"', $table));
```
stonebuzz, line 131: Same fix applies to the orphaned line at line 131:

```suggestion
                $output->writeln(sprintf(__('-> "%s"', 'fields'), $table));
```

inc/container.class.php
```diff
@@ -928,6 +977,175 @@ public static function getTypeName($nb = 0)
...
+        /** @var DBmysql $DB */
+        global $DB;
+
+        $container = new self();
+        if (!$container->getFromDB($id)) {
+            Session::AddMessageAfterRedirect(sprintf(__('Unknown container #%d.', 'fields'), $id), false, ERROR);
+
+            return false;
+        }
+
```
stonebuzz, line 1011: `renameOversizedContainer` has no server-side guard that the target container is actually disabled or oversized.

```suggestion
        if ((int) $container->fields['is_active'] !== 0) {
            Session::AddMessageAfterRedirect(__('Container is already active.', 'fields'), false, ERROR);

            return false;
        }
```

stonebuzz (changes requested):

inc/checkdatabasecommand.class.php
Rom1-B: Nothing to translate

inc/checkdatabasecommand.class.php
Rom1-B: Done

inc/container.class.php
Rom1-B: Good catch, but a strict `is_active === 0` check would break the orphan-recovery path (an active container whose table got corrupted by an old migration bug, see `testRenameOversizedContainerRecoversDataFromOrphanTable`). Added a guard instead that requires the container's *current* name to actually produce an oversized table, same condition already used to show the "Rename and reactivate" button.

PR #1249, opened 2026-08-21
Add psalm

RomainLvr:


Rom1-B (commented):
This PR can target “main” directly, which allows it to be merged quickly, but it will require rebasing the other 12 pending PRs...

RomainLvr:
> This PR can target “main” directly, which allows it to be merged quickly, but it will require rebasing the other 12 pending PRs...

Going back to the main branch won’t make much difference; the changes will be applied in any case once 1247 is merged, which shouldn’t be long now given that the review has been completed and the suggestions have been implemented.
I just wanted to test GitHub’s new Stack system :)

Rom1-B:
Even landing eventually isn't the same as landing as its own commit: stacking it on another PR means it gets squashed into that branch's commit instead of showing up separately in main's history. Could you retarget this to main directly instead of playing with the Stack feature?

RomainLvr:
> Even landing eventually isn't the same as landing as its own commit: stacking it on another PR means it gets squashed into that branch's commit instead of showing up separately in main's history. Could you retarget this to main directly instead of playing with the Stack feature?

I don't think that's how GitHub's new Stacked PR feature works.

The child PR remains a separate PR and can be merged independently. If #1247 is merged first, #1249 is then retargeted onto `main`, and its changes can still be merged separately.

So stacking doesn't mean that #1249 is automatically squashed into #1247's commit or that it disappears from `main`'s history. The final history still depends on the merge method chosen for each PR.

Since #1247 is already reviewed and almost ready to merge, I don't see a downside to keeping #1249 stacked on it. It also lets us review #1249 independently without having to rebase the other 12 pending PRs.

We will have to use the second option for merging :
image

stonebuzz:
Yes please target `main` branch

PR #1254, opened 2026-09-01
fix(container): save default field value when empty on creation

MyvTsv:

## Description

- It fixes !45954
- When a Fields block is embedded in a ticket's base form and contains a required field with a default value, an error may occur when creating a ticket through a collector.

This is because, when the collector created the ticket, the field’s default value was not saved. The required field therefore remained empty, which caused an error during ticket creation.

This PR fixes this behavior by ensuring that the field’s default value is correctly saved when the ticket is created.


Lainow (approved):

RomainLvr (approved):

inc/container.class.php
Rom1-B, line 2149:
```suggestion
                    } elseif ($item->isNewItem()) {
                        $default = PluginFieldsField::getDefaultValue($field);
                        $decoded = json_decode((string) $default, true);
                        if (is_array($decoded) && $decoded !== []) {
                            $data[$field['name']] = $decoded;
                            $has_fields           = true;
                        } else {
                            $data[$field['name']] = [];
                        }
                    } else { //multi dropdown is empty or has been emptied
```

PR #1255, opened 2026-09-03
FIX / Admin locks himself out of a container's configuration

Mary-Clb:

## Description

- It fixes !46060
- On the container's **Profiles** tab, setting the currently active profile's
right to *No access* made the whole container **configuration form**
(`front/container.form.php?id=X`) return "Access denied" — not just the block
on the target itemtype. 
Gate the display branch with `$container->check(id, READ)` (fallback `-1` for
the add form), like `front/container.php` and the other sub-forms. Since
`$rightname = 'config'`, this enforces the standard `config` right + entity
scoping. Removed the now-unused `AccessDeniedHttpException` import.


Rom1-B (approved):
Please rebase

